PR #76, opened 2016-09-08
[install guide.md] Is joining insider program required for using WSL?

xiaoyinl:
This patch removes repeated prerequisites. Also are non-insider users with anniversary update able to use Bash on Windows? I remove it in this patch, but I can add it back if joining insider program is still required.


xiaoyinl:
Hi @bitcrazed and @russalex, could you review this patch? Thanks!

bitcrazed:
Thanks for your change - great clean-up ;)

PR #155, opened 2017-12-09
fixed typos!

slightlytrue:
Hey there! 👋 
I think these typos should be corrected so that they don't confuse other users. 😄


scooley:
Thank you!

PR #156, opened 2017-12-21
Updated interop doc to FCU

scooley:
Added WSLENV and tested/updated all bash -c references to WSL.exe.

### DIFF
--- a/WSL/interop.md
+++ b/WSL/interop.md
@@ -157,7 +157,7 @@ C:\temp>dir | findstr foo.txt
 
 Prior to 17063, only Windows environment variable that WSL could access was `PATH` (so you could launch Win32 executables from under WSL).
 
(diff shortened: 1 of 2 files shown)


WSL/interop.md
```diff
@@ -3,123 +3,183 @@ title: Windows interoperability with Linux
...
 ```
 
-Users may disable the ability to run Windows binaries for a single WLS session by running the following command as root:  
+## Share environment variables between Windows and WSL
+
+> Available in Windows Insider builds 17063 and later.
+
+Prior to 17063, only Windows environment variable that WSL could access was `PATH` (so you could launch Win32 executables from under WSL).
+
+Starting in 17063, WSL and Windows share `WSLPATH`, a special environment variable created to bridge Windows and Linux distros running on WSL.
```
cwilhit: This needs to be changed to 'WSLENV' @scooley

scooley: Fixed!  Good catch.

PR #162, opened 2018-01-21
Fixed a typo in wsl-config.md

thatging3rkid:
The command was `wsl /list` when it should be `wslconfig /list`.


sunilmut:
Thanks, @thatging3rkid

PR #168, opened 2018-03-10
Add link to Debian GNU/Linux in Microsoft Store

AlanMcBee:


leandrobsilva:
Similar a ticket 167.

AlanMcBee:
@leandrobsilva Agreed. Scott's is the better PR. Use 167 instead of this. Do I select "Close and comment"? Still new to githubbing.

scooley:
Reverting and grabbing Scotts because it has Kali too

PR #176, opened 2018-03-22
Update wsl-config.md to remove bad copy pasta

diddledan:
It appears that wsl-config.md might be a reworked version of documentation for Hyper-V as there was an additional unrelated-to-WSL code snippet detailing usage of the `Get-VM` cmdlet. This commit removes that extra code snippet.


diddledan:
Oh, I might have mistaken the intention of that example. It looks now that I recheck everything that it is an example of using `wsl.exe` to call `grep` on the output from `Get-VM`. In that case I'm closing this PR.

PR #202, opened 2018-05-24
Assuming that was a typo and not a future change

jtgasper3:


scooley:
Good catch :).  Updated this already before noticing the PR

PR #225, opened 2018-07-09
Provide instructions on port passthrough

imcdnzl:
Fairly commonly asked question on the web so would be great to add to the FAQ


bitcrazed:
Thanks. Good additional point :)

PR #227, opened 2018-07-12
fixing broken link to Microsoft Store

dnielsen:
also changing name from Windows Store to Microsoft Store


bitcrazed:
Thanks for your suggested edits, but we're not trying to take the user to the (retail) Microsoft Store - we're providing a link to the Linux distros available for download from the Windows Store.

PR #239, opened 2018-07-30
Remove a dangling 'The' from Installation Guide

aishpant:


bitcrazed:
Thanks for the fix :)

PR #255, opened 2018-08-27
Replace Windows Store with Microsoft Store

alexandair:
Replace "Windows Store" with "Microsoft Store".
Replace "command-line" with "command line" when it's used as a noun and not as an adjective.


tara-raj:
Thanks for submitting these changes! Looks good - going to merge

PR #272, opened 2018-09-22
Fix log message rendering

Lin-Buo-Ren:
Signed-off-by: 林博仁(Buo-ren, Lin) <Buo.Ren.Lin@gmail.com>


mscraigloewen:

PR #283, opened 2018-10-10
Update example curl link

sapir:


mscraigloewen:
Hi @sapir I'm closing this without merging since 1604 is still active and a distribution that users may want to use. Thank you for your submission!

sapir:
Yes, but shouldn't the default be 1804?

mscraigloewen:
The string there is just an example, and I don't believe it's necessary for it to reflect the default latest version of Ubuntu.

PR #284, opened 2018-10-12
Make elevated permissions warning flow better

bplubell:
This makes the start of the warning about elevated processes a easier to read. When first reading through the page, I was tripped up with some odd syntax (particularly "that").

The whole warning now has quite a few slashes, but I think it works.

Not sure which branch to submit PR against, but it looks like "live" is the default - if needed, we can change.


mscraigloewen:

PR #295, opened 2018-10-28
Clarify bullet point

fdmarcin:
Hi,
Previously, "ie" was misspelled, and anyway I think "for example" fits here better.


mscraigloewen:

PR #298, opened 2018-10-30
Update WSL/faq.md

ThrDev:
Add change default mounted drive location reference in FAQ, to point to WSL-Config.


mscraigloewen:
Hi @ThrDev , I'm closing this issue without merging as while the info you've added is very useful it doesn't directly answer the question and we try to keep the answers in the FAQ as short as we can. Thank you for your submission!

PR #311, opened 2018-11-13
Update for new error resolution

cloudmelon:
Update with a very common error resolution ( users install WSL directly from windows store and ignore the activation process from control panel or powershell


mscraigloewen:

PR #345, opened 2019-01-28
Update install-win10.md

rrmistry:
Removed duplicated word


mscraigloewen:
This has already been fixed, thanks for your submission!

PR #347, opened 2019-01-31
Fix typo: ie => or

doggy8088:


mscraigloewen:
Hi @doggy8088 , I'm closing this issue without merging as the 'or' convention is clearer to me. I can understand how 'ie' could also be clear, but I would rather keep the convention in that section. Thank you for your submission!

doggy8088:
@mscraigloewen I really through you had a typo. I think change this `ie` to `i.e.` might be better. I'm not native speak in English. I'll confuse what's **ie** mean.

mscraigloewen:
I've merged in another change that says 'for example' which I think makes it a lot clearer. :)
image

PR #363, opened 2019-02-26
Updated commands for Ubuntu 18.04

shibut12:
Instructions to reset password for Ubuntu is outdated, it does not work for latest Ubuntu 18.04. I Updated documentation to reflect this change.


mscraigloewen:
Hi @shibut12 , thank you for your submission! I'm closing this pull request without merge as the regular `ubuntu.exe` is left in as an example. Any distro can have any executable name so we don't want to enumerate more than just one example. Thanks again!

PR #429, opened 2019-06-20
Renames title and recommends wsl.exe

WSLUser:
This doc is still needed as is for developers running versions of Windows 10 under 1803. This simply makes that more clearly known so developers who are on newer versions know to ignore this doc and use the recommended method. Since this doc needs to be maintained, renaming it will be helpful so that a new page can be created that shows the supported procedure.


mscraigloewen:
I'll be updating this whole section tomorrow, so I'm going to close this issue as I'll be reformatting the entire reference page. Thanks for the contribution though!

PR #455, opened 2019-07-13
Correcting python to Node.js

tkore:
> The picture below shows an example of this by connecting to a python server running in Windows via curl. 

The picture following this sentence shows a Node.js script being executed on Windows and a curl request being sent to it, but the line above says the server is in fact in Python.

Also, correcting nodeJS to Node.js in the following line -
> The picture below shows an example of this by connecting to a nodeJS server using the Edge browser.


mscraigloewen:
Great catch! Thank you for the submission.

PR #466, opened 2019-07-30
[Typo] Corrected `%UserProfile%\.wslconfig`

nunix:
there is a missing escape character for displaying the `\`.
The preview shows `Introduce %UserProfile%.wslconfig file for tweaking WSL2 settings` instead of `Introduce %UserProfile%\\.wslconfig file for tweaking WSL2 settings`

Therefore, if someone tries to copy the path in explorer, there will be an error


mscraigloewen:
Thank you for fixing this!

PR #498, opened 2019-10-04
Update about.md

jdorsay-xenu:
UserVoice portal](https://wpdev.uservoice.com/forums/266908-command-prompt-console-bash-on-ubuntu-on-windo/category/161892-bash) appears to be gone.


craigloewen-msft:
Merged! Thank you :)

Issue #517, opened 2019-11-05
both powershell commands require reboots 


christiankuhtz:
both powershell commands above to enable WSL & vm platform require reboots. would be good to break them out into separate copy pastes.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b19a43be-b16a-d903-b69d-75dffce6c329
* Version Independent ID: f9d4a118-1251-8d6b-43f5-2e64808dc527
* Content: [Install WSL 2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install?source=docs#feedback)
* Content Source: [WSL/wsl2-install.md](https://github.com/MicrosoftDocs/WSL/blob/live/WSL/wsl2-install.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

MateuszZiolek:
Hi 
in my case all was done in one round without any reboot, also both WSL machines was converted also without reboot.
Done at fresh Version 10.0.19018.1

christiankuhtz:
Sure. But that doesn't prove anything other than this appears to be not consistent.  Given that I was prompted twice to reboot, that copy paste should be broken up into two in this doc. ;-)

MateuszZiolek:
well i have already WSL v1 two units so maybe this was a case with do not trigger the reboot.

christiankuhtz:
Yup. That's why I'm suggesting this be broken out into two fields that can be copy pasted.

craigloewen-msft:
I made these changes, it's now live: 610d9aa

Thank you for filing this! :)

PR #518, opened 2019-11-06
break out into two copy pasteable fields

christiankuhtz:
given that it's not guaranteed that each of the commands don't require a reboot by themselves, this needs to be broken out into two copy pasteable instructions. #517


craigloewen-msft:
Thank you for adding this into a PR. Unfortunately I only just saw this after I rewrote the section, but I have pulled out the two commands into separate fields and added some more context about restarting, thank you for helping improve our docs! :)

PR #525, opened 2019-11-19
Update install-win10.md

taddieken95:
An ordered list is confusing to the end user when presenting choices. I'd advocate for using unordered bullets so that the user doesn't think these options are a procedural order.


craigloewen-msft:
Thanks for the suggestion! Merged :)

taddieken95:
woot!

PR #531, opened 2019-12-05
update instructions to include slow ring

sanyides:
WSL2 is now available for the 'Slow' ring as well but the documentation says you need to use the fast ring.


craigloewen-msft:
Great fix! Thank you for making this PR, it's merged. :)

PR #546, opened 2020-01-13
Moved instructions for WSL access in Windows

jpopadak:
- Made newer builds more prominent as 20H1 release is soon


craigloewen-msft:
Great addition! Thanks for the PR :)

PR #549, opened 2020-01-14
Docs staging

craigloewen-msft:
Adding changes from staging docs

### DIFF
--- a/WSL/user-support.md
+++ b/WSL/user-support.md
@@ -1,137 +1,53 @@
 ---
-title: Linux User Account and Permissions
+title: Create and update user accounts for WSL distributions
 description: Reference for user accounts and permission management with the Windows Subsystem for Linux.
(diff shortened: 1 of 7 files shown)


WSL/user-support.md
```diff
@@ -1,137 +1,53 @@
...
+### Forgot your password
 
-    ```console
-    C:\> lxrun /setdefaultuser root
-    ```
+If you forgot the password for your Linux distribution, try this: 
 
-1. Run `bash.exe` to now login as `root`:
+For Ubuntu:
+1. Open PowerShell and use it to change the default user name of your Ubuntu distribution to root with: `ubuntu config --default-user root`
```
craigloewen-msft: You don't need to do this change default user business. You can actually run as root directly!
`wsl -u root` will start you off in your default distro as root, and `wsl -d Debian -u root` will run you as root in the Debian distro.

From there you can go in and run the `passwd` command and you won't need to reset your default user after that. :) 

This should simplify this whole section!

PR #550, opened 2020-01-14
Rewrite of troubleshoot pg draft 1

mattwojo:


craigloewen-msft:
I added in some suggested changes in the form of a new commit. I think the wording still needs work, but what do you think of the content changes?

I moved everything around slightly, only bit I removed was the Docs feedback instructions since they're at the bottom of every docs page anyway!

craigloewen-msft:
Looks good!

Let's merge this into docs staging if it looks good to you, and then we can merge docs staging into live and delete the branches.

mattwojo:
Still some fixes needed -- we don't want to initially push folks to https://github.com/microosft/wsl/issues, do we?  ...wouldn't we want to first send them to:  https://github.com/MicrosoftDocs/wsl/issues?  ...my initial purpose was to clarify that both exist and the difference between them...

craigloewen-msft:
> Still some fixes needed -- we don't want to initially push folks to https://github.com/microosft/wsl/issues, do we? ...wouldn't we want to first send them to: https://github.com/MicrosoftDocs/wsl/issues? ...my initial purpose was to clarify that both exist and the difference between them...

For any technical issue, users should go straight to the Microsoft/WSL repo. The MicrosoftDocs/WSL repo should only be for issues relating specifically to our documentation (things like, this page is confusing, or has a spelling mistake).

I'd rather push users to go straight to Microsoft/WSL since it's the most active, and if we do get the occasional false positive of docs questions going there then that's fine by me.

PR #580, opened 2020-03-04
Activate links to Ubuntu 18.04

dev-bittlinger:
Links were already added correctly but behind a comment without direct reasoning. I checked both and they work, so I assume we can activate them again, for easing the use for the end-user.


dev-bittlinger:
fixes #576

Biswa96:
Those links were removed in very previous commit without any reason.

dev-bittlinger:
This got fixed by commit 0082a53

Issue #594, opened 2020-03-16
Incorrect Platform: AMD64  instead of x64


puhazh:
The download link points to x64 but the text mentions it as AMD64. There is a separate note section below for AMD64
Text:
Please click on this link to download the latest WSL2 Linux kernel update package for AMD64 machines.

Should be:
Please click on this link to download the latest WSL2 Linux kernel update package for x64 machines.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5d720f69-aff2-38d8-f0c6-9909e94acaa2
* Version Independent ID: 9d121625-e741-6d20-f3de-201a939290b8
* Content: [Updating the WSL 2 Linux kernel](https://docs.microsoft.com/en-us/windows/wsl/wsl2-kernel#feedback)
* Content Source: [WSL/wsl2-kernel.md](https://github.com/MicrosoftDocs/WSL/blob/live/WSL/wsl2-kernel.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

PR #595, opened 2020-03-16
Issue #594: Incorrect Platform AMD64 instead of x64

puhazh:
The download link points to x64 but the text mentions it as AMD64. There is a separate note section below for AMD64


craigloewen-msft:
Great spot! Merged. Thanks for your contribution! :)

puhazh:
Closes #594

PR #603, opened 2020-03-23
fix invalid windows version check command

alyahmedaly:


mattwojo:
Thanks for contributing @alyahmedaly!  The `ver` command actually does work for identifying your Windows version... but only when used in Windows Command Prompt... you're correct that it's `winver` in PowerShell. We just overhauled the WSL docs and this is more specific now in the install document. Thanks for your help. :)

PR #606, opened 2020-03-24
Added section for changing default user

misczak:
Added section for changing default user


mattwojo:
Thanks for your contribution @misczak ! :) merged!

PR #636, opened 2020-04-18
Update Linux distribution list

loganrosen:
I've updated this list based on what's available in the Microsoft Store.


mattwojo:
Thanks @loganrosen ! Appreciate your contribution! :)

PR #639, opened 2020-04-20
refine FAQ about backup to cover move of distros also

carehart:
I had looked to this page previously for info on how to move distros among drives, but I had not noticed that the answer was implied in the discussion of backing up. So I am proposing here a slight refinement to the question and the answer, to make those points stand out better. This is important for those who may come looking for info when their drive is filling and WSL (the distros and/or their apps) are a culprit.


mattwojo:
Thanks @carehart ! Appreciate your contribution! :)

PR #644, opened 2020-04-23
Specify administrator PS window for WSL2 install

schultetwin1:
Specify that users running the dism.exe install commands should do so in an administrator window


mattwojo:
Thanks for pointing this out @schultetwin1 ! 
We've deprecated the wsl2-install page, choosing instead to combine it with WSL 1, as it's really a single install and then an update to WSL 2 currently. I'm going to make sure that opening PowerShell as an admin is included. Thanks again.

PR #668, opened 2020-05-11
Add link to Ubuntu and Ubuntu 20.04 LTS

patrick330602:
Ubuntu 20.04 LTS is released for a while, and I found that link to Ubuntu is missing in the Installation page, thus adding them.


mattwojo:
Thanks @patrick330602!

PR #670, opened 2020-05-11
Combine compare and whats new, add redirect

mattwojo:


mattwojo:
DEcided against combining these pages.

PR #689, opened 2020-05-20
Fix typo in install-win10.md

nicholasz2510:


mattwojo:
Thanks for the typo fix @nicholasz2510! Good catch. :)

PR #739, opened 2020-06-10
Fix bullet point formatting on About page

renodubois:
These were missing spaces and weren't getting displayed correctly in the docs website.


mattwojo:
Thanks @renodubois!

PR #752, opened 2020-06-16
Update VMware / VirtualBox Compatibility

HeroCC:
VMWare workstation released version 15.5.5 with support for Hyper-V, so I updated the version and link in the alert. I also updated the chart to reflect that the latest versions of VMware and VirtualBox both support running alongside WSL2 / HyperV.


craigloewen-msft:
Great addition! Thank you for adding it :)

PR #765, opened 2020-06-21
Added a clickable link to the MSI kernel update

haakoo:
Why make it harder than necessary to get the MSI-package?


craigloewen-msft:
Great addition! Thank you. :)

PR #781, opened 2020-06-29
Include cmdlets dism

EwertonJordao:
Include Enable-WindowsOptionalFeature, dism module PowerShell I think is more clear to read cmdlet then dism command.

### DIFF
--- a/WSL/install-win10.md
+++ b/WSL/install-win10.md
@@ -18,6 +18,10 @@ Open PowerShell as Administrator and run:
 ```powershell
 dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart
 ```
+or 
+```powershell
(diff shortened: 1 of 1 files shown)


mattwojo:
Hey @craigloewen-msft - Could you review and either approve or close? I don't know enough about the differences between these two PowerShell scripts.

WSLUser:
They both will work. The header makes it confusing but the file changes make it obvious. They do the same thing but dism is a cmd utility that can run in powershell while the other method is powershell specific.

EwertonJordao:
> 
> 
> They both will work. The header makes it confusing but the file changes make it obvious. They do the same thing but dism is a cmd utility that can run in powershell while the other method is powershell specific.


Yes, they do the same task. I think include that step on the documentation make it easer to undestend whats happen

mattwojo:
My opinion is that we pick one of these scripts, including both feels like it could potentially add to confusion for new users. Let's be opinionated. If this new script makes it easier to understand what's happening, then it seems like a better option, but I'd like @craigloewen-msft to make the decision as the PM.

PR #789, opened 2020-07-08
Push live get started with databases tutorial

mattwojo:
...and set-default note regarding os version needed


mattwojo:
Updated title to 'Archived' for old instructions, added noindex metadata, and removed Loc metadata tag.

PR #794, opened 2020-07-13
Update win10

EwertonJordao:
Include a script for update distros and install wsl_update_x64.

### DIFF
--- a/WSL/install-win10.md
+++ b/WSL/install-win10.md
@@ -113,6 +113,47 @@ wsl --set-default-version 2
 
 This will set the version of any new distribution installed to WSL 2.
 
+## Install and upgrade all distros 
+
+Run the following command in PowerShell for upgrade all your distros and install wsl update package.
+```powershell
+#WSL2 Install and Update
+#enable feature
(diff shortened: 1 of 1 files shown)


mattwojo:
This is pretty slick. @craigloewen-msft - What do you think about adding a section with Ewerton's PowerShell update script?

WSL/install-win10.md
```diff
@@ -113,6 +113,47 @@ wsl --set-default-version 2
 
 This will set the version of any new distribution installed to WSL 2.
 
+## Install and upgrade all distros 
+
+Run the following command in PowerShell for upgrade all your distros and install wsl update package.
```
dimithras, line 118: I guess better wording will be:
- run (...) **to** ungrade
or
- run (...) for initiating and upgrade of all...

Otherwise does not sound right.

WSL/install-win10.md
```diff
@@ -113,6 +113,47 @@ wsl --set-default-version 2
 
 This will set the version of any new distribution installed to WSL 2.
 
+## Install and upgrade all distros 
+
+Run the following command in PowerShell to upgrade all your distros and install wsl update package.
```
EwertonJordao, line 118: Include **to**

PR #805, opened 2020-07-21
embed what can i do with wsl video

mattwojo:
..based on dev env user study feedback wanting a short video to explain what wsl is


craigloewen-msft:
👍 We'll need to fix the merge conflicts, but otherwise ship it!

PR #814, opened 2020-07-25
Update user-suport.md

Himanshunitrr:


mattwojo:
Tweaked this a bit to maintain signularity but also be more specific that the user name and password is distinct for each Linux distro installed.

PR #818, opened 2020-07-30
Use correct capitalization of VMware

geary:


mattwojo:
Doh! Thanks for the correction @geary !

PR #820, opened 2020-07-31
Update wsl-config.md: Fix `memory` default

galeksandrp:
This commit reflect change in default memory consumption introduced in `20175` build

https://docs.microsoft.com/windows/wsl/release-notes#build-20175


mattwojo:
Ahh! Great catch @galeksandrp. Thank you for the update!

PR #848, opened 2020-08-19
Add 'command not found' to troubleshooting page

licanhua:
Address issue https://github.com/microsoft/WSL/issues/5779


WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
 
 ## Common issues
 
+### `command not found` when execute windows .exe in linux
```
licanhua, line 39:
```suggestion
### `command not found` when executing windows .exe in linux
```

therealkenc:
> PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH"

That will cause (for example) `/bin` to be in the final path twice.

[Tangential trivia. The `id -u` test presumes `/usr/bin` is in the `$PATH` before `$PATH` is set.]

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
 
 ## Common issues
 
+### `command not found` when executing windows .exe in linux
+User can run Windows .exe like notepad.exe directly from linux. Sometime, you may hit "command not found" like below:
```
craigloewen-msft, line 40:
```suggestion
User can run Windows executables like notepad.exe directly from linux. Sometimes, you may hit "command not found" like below: 
```

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
 
 ## Common issues
 
+### `command not found` when executing windows .exe in linux
+User can run Windows executables like notepad.exe directly from linux. Sometimes, you may hit "command not found" like below:
```
craigloewen-msft, line 40:
```suggestion
Users can run Windows executables like notepad.exe directly from Linux. Sometimes, you may hit "command not found" like below: 
```

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
...
 ## Common issues
 
+### `command not found` when executing windows .exe in linux
+User can run Windows executables like notepad.exe directly from linux. Sometimes, you may hit "command not found" like below: 
+```
+$ notepad.exe
+-bash: notepad.exe: command not found
+```
+
+If there is no win32 paths in your $PATH, interop isn't going to find the .exe.
```
craigloewen-msft, line 46:
```suggestion
If there are no win32 paths in your $PATH, interop isn't going to find the .exe.
```

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
...
+You can verify it by `echo $PATH` in Linux. It's expected that win32 path(for example, /mnt/c/Windows) is in the output.
```
craigloewen-msft, line 47:
```suggestion
You can verify it by running `echo $PATH` in Linux. It's expected that you will see a win32 path (for example, /mnt/c/Windows) in the output.
```

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
...
+If you can't find it, most likely PATH is overwriten by Linux shell.
```
craigloewen-msft, line 48:
```suggestion
If you can't see any Windows paths then most likely your PATH is being overwritten by your Linux shell. 
```

WSL/troubleshooting.md
```diff
@@ -36,6 +36,39 @@ Lastly, if your issue is related to the Windows Terminal, Windows Console, or th
...
+```
+if [ "`id -u`" -eq 0 ]; then
+  PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+else
+  PATH="/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"
+fi
+```
+
+The correct way on Debian is to remove above lines.
+You may also append $PATH during the assigment like below, but this lead to some [other problems](https://salsa.debian.org/debian/WSL/-/commit/7611edba482fd0b3f67143aa0fc1e2cc1d4100a6) with WSL and VSCode..
```
craigloewen-msft, line 60: Can you remove the lines after this? We should give them one clear solution, which in this case is should be just to remove the lines of the shell that sets the Path. Once done we can merge this in!

therealkenc:
If this doc update is going to refer to an issue, please make it WSL# 5779 not 5296. The OP in 5779 is on Ubuntu not Debian, and Ubuntu doesn't suffer the `/etc/profile` problem. 5779 is presently on a slow-path to being question-closed, answer-shrug, absent some kind of follow-up. 5296 is an understood problem tagged external. It was a known regress caused by change to the WSL Debian launcher, with the work-around of deleting those few lines in `/etc/profile` described.

mattwojo:
@craigloewen-msft - this looks good to me, could you confirm and merge?

@licanhua - Thank you for this conribution! I think it will help a lot of folks.

PR #860, opened 2020-08-24
Add link to WSL 2 kernel update page in left-hand table of contents

metathinker:
Also corrected a minor typo in that navigation pane.

### DIFF
--- a/WSL/toc.yml
+++ b/WSL/toc.yml
@@ -4,7 +4,7 @@
   items:
     - name: What is WSL?
       href: about.md
-    - name: Whats new in WSL 2 
+    - name: What's new in WSL 2?
       href: wsl2-index.md
     - name: Comparing WSL 2 and WSL 1
       href: compare-versions.md
@@ -38,6 +38,8 @@
       href: enterprise.md
     - name: Build a custom distro
       href: build-custom-distro.md
+    - name: Update WSL 2's Linux kernel
+      href: wsl2-kernel.md
     - name: Command Line Reference
       href: reference.md
 - name: Frequently Asked Questions


metathinker:
ping @craigloewen-msft

craigloewen-msft:
Adding the kernel install page to the TOC may not be the best solution here. In the future we want this step to be invisible, and in the latest insiders preview builds when you install WSL using `wsl --install` it installs the kernel automatically for you, meaning you don't have to worry about adding the kernel yourself, and so having it in the TOC to install the kernel could be confusing for some users.

However, you do raise a good point that we aren't highlighting this step enough now as it is. We have this PR in the works to make the kernel install step explicit in the install docs. Does this help resolve your feedback?

https://github.com/MicrosoftDocs/WSL/pull/912/files

metathinker:
> Adding the kernel install page to the TOC may not be the best solution here. In the future we want this step to be invisible, and in the latest insiders preview builds when you install WSL using `wsl --install` it installs the kernel automatically for you, meaning you don't have to worry about adding the kernel yourself, and so having it in the TOC to install the kernel could be confusing for some users.

Is there an intention to backport this `wsl --install` improvement back to 1903 and up via servicing? If not, there are still going to be a non-zero number of users who have to install the WSL 2 kernel manually, and it might be easier to search the web for this page instead of having to remember the URL shown in the prompt https://aka.ms/wsl2kernel . My hope is that with this link the right page will show up higher in web searches.

> However, you do raise a good point that we aren't highlighting this step enough now as it is. We have this PR in the works to make the kernel install step explicit in the install docs. Does this help resolve your feedback?
> 
> https://github.com/MicrosoftDocs/WSL/pull/912/files

I think you might want to sync with the MSFT-internal folks who were raising this concern originally (I think @adegeo was one of them?). They might be better able to comment on whether that other PR better meets their needs.

mattwojo:
@metathinker - thanks for contributing! Agree that this was a bit confusing.
I rewrote the kernel install instructions (https://github.com/MicrosoftDocs/WSL/pull/916/files), incorporating them into the main WSL install page, so I'm going to close this out. As Craig pointed out, the WSL install story is continuing to change (and hopefully become much easier and more straight-forward), but this seemed to best way to me to handle the current install state. Please let me know if you don't agree or find anything confusing. Thanks again for contributing!!

mattwojo:
I guess I spread changes over a couple of PRs... but you can see the results best on the live site: https://docs.microsoft.com/windows/wsl/install-win10#step-4---download-the-linux-kernel-update-package

PR #881, opened 2020-08-31
wsl –set-version with single dash is faulty

dimithras:
Single dash `wsl –set-version` is not going to work as syntax for this option must have double dash. The right one: `wsl --set-version`.

> --set-version \<Distro\> \<Version\>
Changes the version of the specified distribution.


craigloewen-msft:
Thank you for the fix!

PR #885, opened 2020-09-01
Fix minor typo

pablojimpas:
Extra backticks removed


craigloewen-msft:
Thank you for the fix!

PR #896, opened 2020-09-09
Add docker + wsl 2 tutorial doc

mattwojo:


mattwojo:
@craigloewen-msft Does everything look ok here? Did you have any other feedback you'd like to incorporate before merging into master branch?

PR #900, opened 2020-09-10
Fix a typo

alexandair:
Fix a typo.


mattwojo:
Doh! Good catch @alexandair

PR #901, opened 2020-09-10
Update release-notes.md for 20211

benhillis:


mattwojo (approved):
>Updated the link text to match the article, rather than using "here"
>Added periods to make the punctuation consistent

PR #917, opened 2020-09-17
Removed dead link in the build-custom-distro.md file

BeardedFish:
The removed link ("Command-line UserVoice portal") leads to a 404 page that says: "This UserVoice instance is no longer available."


mattwojo:
Great catch. Thanks @BeardedFish! We dropped support for UserVoice quite a while back in favor of using GitHub issues. Thanks again.

PR #936, opened 2020-09-25
Add enable steps for gui rather than script

mattwojo:


mattwojo:
Closing - PM prefers to keep streamlined with just PS script and avoid more noise / steps from this GUI option.

PR #942, opened 2020-10-01
Update install-win10.md

mikloscari:
just add more clarity that Storage is 'inside' System


mattwojo:
Thanks @mikloscari : )

PR #948, opened 2020-10-05
Links: WSL (2020-10)

DCtheGeek:
@mattwojo Monthly FQDN cleanup for WSL.


mattwojo:
Thanks @DCtheGeek. Both of these links were actually related to a page that we're phasing out -- keeping the page temporarily (while doing some Loc testing), but replaced with the link to the new location for the kernel instructions.

PR #949, opened 2020-10-05
Update virtualization not enabled error section

veleek:
Updating section describing what to do when virtualization not enabled to include error message and error code.


craigloewen-msft:
Great change! Thank you for the contribution.

PR #961, opened 2020-10-16
Adding more emphasis one when to Restart through the process

espipj:


mattwojo:
More emphasis it is @espipj! :) Thanks.

PR #984, opened 2020-11-04
Updated wsl2-mount-disk.md to include Mount-VHD instructions

craigloewen-msft:


benhillis (approved):
Agreed with Pierre's comments.

PR #986, opened 2020-11-04
metadata is disabled by default

therealkenc:


mattwojo:
Great catch! Thanks @therealkenc !! <3

PR #987, opened 2020-11-05
Minor Mount updates

mattwojo:


Biswa96:
Tip: There are many typo fixing is going on. _If possible_, use [Notepad++][1], it has a [spell check plugin][2].

[1]: https://notepad-plus-plus.org/downloads/
[2]: https://github.com/Predelnik/DSpellCheck

mattwojo:
Thanks @Biswa96. I think that this was published during a bit of busy time... we typically do better spell check... we generally recommend [VS Code](https://code.visualstudio.com/download) though... in addition to spell check, as well as an [authoring pack](https://marketplace.visualstudio.com/items?itemName=docsmsft.docs-authoring-pack) with lots more goodies specifically to support authoring on docs.microsoft. : ) 
I do still appreciate the simplicity and versitility of Notepad++ though.

PR #999, opened 2020-11-19
fix bad URL

stratusjerry:
fixes a bad URL


alivx (approved):

mattwojo:
Thanks for catching that @stratusjerry!!

PR #1005, opened 2020-11-24
Update wsl-containers.md

eyalb-proftit:
in the comment, a "$" sign is missing


mattwojo:
Sorry for my slow reply @eyalb-proftit. Appreciate your contribution! ...but, both \\wsl\ and \\wsl$\ will work. Started with \\wsl$\, but changed it to \\wsl\... so this update isn't necessary any longer, but please keep the contributions coming if you see anything else!

PR #1008, opened 2020-11-27
code sample for 'append $PATH during the assignment like below'

wswind:


mattwojo:
Thanks for the contribution @wswind ! Sorry for the delay in merging.

PR #1010, opened 2020-11-28
Update compare-versions.md

tarikku:
Serial and USB access are further use cases for WSL over WSL 2


mattwojo:
Great addition. Thanks for contributing this @tarikku!! Very useful to point out!

PR #1016, opened 2020-12-03
Initial --install fixes

craigloewen-msft:
Fixes to make --install more prominent.

Authored by @bpulliam


craigloewen-msft:
Added a commit:

Removed image
Moved [Note] up to make it more visible
Reverted the message of 'Choose WSL 2 or WSL 1' back to 'Update to WSL 2'

Issue #1042, opened 2021-01-04
WSL 2 installation doc: note placed incorrectly in the document.


ElmarJ:
The "Note" that starts with "The update from WSL 1 to WSL 2 may take several minutes" is not placed logically in the document. It is placed under section "Step 5 - Set WSL 2 as your default version", but I would have expected it under section "Set your distribution version to WSL 1 or WSL 2".

This caused some confusion, as I expected the conversion process to start with the command mentioned under "Step 5".

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b7a1833a-3461-2d9e-bb1e-de5d0ee3fe33
* Version Independent ID: bd4b2627-60c7-1bfe-7831-1dbde906e287
* Content: [Install Windows Subsystem for Linux (WSL) on Windows 10](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
* Content Source: [WSL/install-win10.md](https://github.com/MicrosoftDocs/wsl/blob/master/WSL/install-win10.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

craigloewen-msft:
Agreed! This is a great change. I've merged your PR, thank you for the contribution!

PR #1043, opened 2021-01-04
WSL-installation doc: corrected position of note

ElmarJ:
Corrected note about conversion-time of WSL-upgrade, which was placed in the wrong section.

Solves #1042


craigloewen-msft:
Agreed! Thank you for the contribution. :)

PR #1045, opened 2021-01-04
Update wsl-git.md

rhorber:
Fix wording. "home" and "root" got mixed up.


mattwojo:
Great catch! Thank you @rhorber!

PR #1065, opened 2021-01-20
Remove description about ver command

tzccinct:
> To check your version and build number, select Windows logo key + R, type winver, select OK. (Or enter the ver command in Windows Command Prompt).

The `ver` command in Windows Command Prompt actually does not show necessary version number (Release ID).
```
> ver

Microsoft Windows [Version 10.0.19041.746]
```

To check Release ID in command line, you need the following command in PowerShell:
> PS > (Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ReleaseId

It's awkward and it's easier to use `winver`. I suggest to remove the description about `ver` command.


mattwojo:
Thanks for pointing this out and the fix contribution @tzccinct!

PR #1068, opened 2021-01-25
install-win10.md: add caveat concerning search expression

kriegaex:
On non-English Windows versions, you might have to modify he search text, e.g. in German it would be `systeminfo | find "Systemtyp"`.


mattwojo:
Thanks @kriegaex!

PR #1073, opened 2021-02-02
Improve mount limitations documentation

guillermo:
Mounting disk or partitions belonging to a disk that is already in use is not an option and is the cause of several issues: #5997 #6015.

This patch updates the documentation to inform about the limitations of the command, preventing future users from discovering this too late.

### DIFF
--- a/WSL/wsl2-mount-disk.md
+++ b/WSL/wsl2-mount-disk.md
@@ -2,7 +2,7 @@
 title: Get started mounting a Linux disk in WSL 2 (preview)
 description: Learn how to set up a disk mount in WSL 2 and how to access it.
 keywords: wsl, windows, windowssubsystem, gnu, linux, bash, disk, ext4, filesystem, mount
-ms.date: 11/04/2020
+ms.date: 04/08/2021
 ms.topic: article
 ms.localizationpriority: medium
 ---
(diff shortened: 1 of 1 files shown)


guillermo:
After a comment from @OneBlue in https://github.com/microsoft/WSL/issues/6015#issuecomment-772040335 I realize that the documentation informs about the limitations at the button of the page in the **limitations** section.

It felt for me as a step by a step guide. This kind of guide has limitations, and disclaimers at the beginning, from that point of view, for me this commit still makes sense. But I don't know how wide is this interpretation.

craigloewen-msft:
@OneBlue please sign off and merge if the info looks good here!

WSL/wsl2-mount-disk.md
```diff
@@ -16,6 +16,7 @@ This tutorial will cover the steps to identify the disk and partition to attach
 > [!NOTE]
 > You will need to be on Windows 10 Build 20211 or higher to access this feature. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
 > Administrator access is required to attach a disk to WSL 2.
+> WSL 2 `mount` command does not support mounting disk (or partitions belong to a disk) that is currently in use (You can't mount a partition from the same disk as Windows).
```
OneBlue, line 19: >  (or partitions belong to a disk)
This is a bit misleading because `wsl --mount` always attaches the entire disk even if only a partition is requested.

Let's remove that first parenthesis. 

> (You can't mount a partition from the same disk as Windows)

Let's be explicit and write it as `You can't mount the Windows installation disk`

OneBlue (commented):
LGTM.

A couple minor changes and let's merge it.

mattwojo:
Merging. Thanks @guillermo!!

PR #1088, opened 2021-02-14
Update install-win10.md

stevenjudd:
Updated the PowerShell commands to install the Windows Subsystem for Linux and the Virtual Machine Platform using the PowerShell cmdlets from the Dism module. We might as well use the native PowerShell functions rather than call the dism.exe executable.


mattwojo:
Thanks @stevenjudd, but there have been issues with using the PowerShell cmdlets: https://github.com/MicrosoftDocs/WSL/search?q=Enable-WindowsOptionalFeature+-FeatureName&type=issues

One reason for using Dism is that it works better for ARM64 users.

PR #1095, opened 2021-02-17
Validation: Fix redirection rule

DCtheGeek:


DCtheGeek:
@mattwojo Here's the other PR to fix the last issue.  The error message was saying that you can't use `redirect_document_id` of `true` on the same `redirect_url` more than once.  There were two entries pointed at `install-win10` with redirect set to true.  If this is alright, go ahead and merge it.  If you wanted the redirect on the document on the other, let me know and I'll swap it. :)

mattwojo:
Looks good. Thanks @DCtheGeek. Yes - I run into this all the time. This file is outdated enough that anything pointing to it should be at least a few years old.

Issue #846, opened 2020-08-18
add explanation for error 0x80370102 that old CPUs without SLAT cannot run WSL2


encip:
I suggest adding to the troubleshooting section an explanation for the following error message that may appear when trying to install a WSL2 distro on a machine with a Core 2 or older processor. (This error message also has other potential causes, such as virtualization disabled in BIOS.)

> WslRegisterDistribution failed with error: 0x80370102
> Error: 0x80370102 The virtual machine could not be started because a required feature is not installed.

The issue is that SLAT, introduced with Nehalem and later processors, is required for WSL2 (though not required for running Windows 2004!). If you have an older CPU, you will still be able to activate the Virtual Machine Platform and WSL windows features (Task Manager will even report Virtualization:Enabled), it's only when you go to install a distro and you have the default set to WSL2 type that you will get stuck with this error.

The workaround is to change your default version back to WSL1 if you must use WSL on these older processors.

See further documentation of this issue here: https://github.com/microsoft/WSL/issues/4709. I experience the issue with a Intel Core 2 Extreme QX9300 on Windows 2004 19041.450.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b7a1833a-3461-2d9e-bb1e-de5d0ee3fe33
* Version Independent ID: bd4b2627-60c7-1bfe-7831-1dbde906e287
* Content: [Install Windows Subsystem for Linux (WSL) on Windows 10](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
* Content Source: [WSL/install-win10.md](https://github.com/MicrosoftDocs/wsl/blob/master/WSL/install-win10.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

craigloewen-msft:
We've added a detailed section on this under troubleshooting. Thank you for filing!

PR #1099, opened 2021-02-18
add additional cause of error 0x80370102 - CPU does not support SLAT

encip:
The following error can occur if the CPU is old enough to not support SLAT: 

"Error: 0x80370102 The virtual machine could not be started because a required feature is not installed."

The error message itself is not specific to the root issue and I think it would be helpful to list this possible cause here.

References:
https://github.com/microsoft/WSL/issues/4709
https://github.com/MicrosoftDocs/WSL/issues/846


mattwojo:
Thanks for bringing this to attention @encip. Appreciate the contribution to the docs!

PR #1109, opened 2021-03-01
Add `case` in the mount option documentation

patrick330602:
`case` option is available (Ref: https://devblogs.microsoft.com/commandline/per-directory-case-sensitivity-and-wsl/) for DrvFs but it is not documented in the mount options section here(https://docs.microsoft.com/en-us/windows/wsl/wsl-config#mount-options)


mattwojo:
Great addition. Thank you @patrick330602! Appreciate it!

PR #1110, opened 2021-03-01
Update install-win10.md

karthikeyanVK:
not working in ps core without the space


Biswa96:
There **must not** be any space.

mattwojo:
Hmm... wondering if you have some unique settings configured or something @karthikeyanVK. The default, out-of-box experience for using PowerShell to run `wsl --set-default-version 2` is no space before the flag dashes.  If you can share a way for me to reproduce the experience you're having, I'll investigate. 
Closing this PR as it's not currently the default behavior for majority of users. Thanks!

PR #1155, opened 2021-04-12
install-win10.md typo fixes

vaughns:
Error section: missing backslash, missing percent char


mattwojo:
This change (from C:\Users\... to C:\Users\\...) is not correct... at least on my machine. Could you include a screenshot or further explanation for why you think this may be these double blackslashes may be correct in your usage..? Closing for now, but please feel welcome to reopen.

Issue #1172, opened 2021-04-30
Ubuntu 16.04 LTS is no longer available from the Microsoft store


HarryGwinnell:
Ubuntu 16.04 should likely be removed from the docs as it is no longer available from the store

![image](https://user-images.githubusercontent.com/10196821/116700952-423be000-a9bf-11eb-907a-19118011e9c7.png)

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0cc9a77f-e6a1-a719-c337-7ff7506a1ac8
* Version Independent ID: 7ed8fb81-5b8f-4d56-7b89-bf600ce6080a
* Content: [Install WSL on Windows 10](https://docs.microsoft.com/en-gb/windows/wsl/install-win10#step-4---download-the-linux-kernel-update-package)
* Content Source: [WSL/install-win10.md](https://github.com/MicrosoftDocs/wsl/blob/master/WSL/install-win10.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

Querynalhabsyi:
Just try the 18 or 20

PR #1173, opened 2021-04-30
Remove Ubuntu 16.04 as no longer available

HarryGwinnell:
Remove the link to Ubuntu 16.04 Windows Store item as it has been removed. I've left the direct link to the appx package as is still working, though I'm not sure I'd recommend 16.04 at this point as it will be EOL as of tomorrow. Can remove as part of this PR if people deem necessary.

Resolves issue #1172


mattwojo:
Thanks @HarryGwinnell - much appreciated!

PR #1181, opened 2021-05-16
Update install-win10.md

bad-samaritan:
Fix broken formatting


mattwojo:
Thanks @bad-samaritan !

PR #1186, opened 2021-05-21
Update wsl-config.md

0xbadfca11:
> case=off is the new default drvfs mount type
> https://docs.microsoft.com/en-us/windows/wsl/release-notes#build-17763-1809

Version 1803 has been end of life, no annotations required.


mattwojo:
Thanks @0xbadfca11 - good catch!

Issue #499, opened 2019-10-05
wsl : The term 'wsl' is not recognized as the name of a cmdlet,


geeksupport1:
Hi,

 I cannot run wsl.

 I have install ubuntu18.02 through windows store and running it.

I am using Window 10 home on a Corei7 4720HQ

OS build 18995.1

My windows version is
`Microsoft Windows [Version 10.0.18995.1]`

I have followed instructions on
https://docs.microsoft.com/en-us/windows/wsl/wsl2-install

When in powershell as admin i type in
```
PS C:\WINDOWS\system32> wsl
wsl : The term 'wsl' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ wsl
+ ~~~
    + CategoryInfo          : ObjectNotFound: (wsl:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

```
I cannot see wsl in the path
`dir C:\Windows\System32\wsl*`

Any hints?

I have tried all the commands in the links.
https://docs.microsoft.com/en-us/windows/wsl/wsl2-install
https://github.com/MicrosoftDocs/WSL/issues/405

Thanks in advance.

---

#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b19a43be-b16a-d903-b69d-75dffce6c329
* Version Independent ID: f9d4a118-1251-8d6b-43f5-2e64808dc527
* Content: [Install WSL 2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-install)
* Content Source: [WSL/wsl2-install.md](https://github.com/MicrosoftDocs/WSL/blob/live/WSL/wsl2-install.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

geeksupport1:
Strange as it is.

wsl -l v works in CMD but not powershell.

So issued is resolved.

wsl is visible in cmd.exe but not in powershell.

Do not know why.

The rest of the commands relating to wsl wslconfig work in cmd but not powershell.

I have restarted the PC many times.

regards

plassa:
I have same issue where wsl works invoked from cmd.exe, but not from powershell

@geeksupport1, why did you close this?

[...]

PR #1206, opened 2021-06-15
Update install-win10.md

Aeprox:
The wsl command doesn't work in powershell, use CMD instead.

See https://github.com/MicrosoftDocs/WSL/issues/499


mattwojo:
Thanks for this contribution @Aeprox. I reviewed the issue in the WSL product repo that you linked, but have decided not to approve this change. If the wsl command isn't working in PowerShell, there is likely another issue that needs to be addressed.

PR #1235, opened 2021-07-22
fix: Added In Preview Options for .wslconfig

Agrover112:
- Added In Preview Options section at Line 354
- Added option  **guiApplications** to turn on/off support for WSL GUI applications
**Note:** Other In Preview Options may be added if present.


Agrover112:
@craigloewen-msft  Cla signed

PR #1256, opened 2021-08-12
Adds case sensitivity doc

mattwojo:


WSL/case-sensitivity.md
```diff
@@ -0,0 +1,181 @@
...
+> For more WSL configuration options, see [Configure per distro launch settings with wslconf](./wsl-config.md#configure-per-distro-launch-settings-with-wslconf).
+
+### Changing the case sensitivity on a drive mounted to a WSL distribution
+
+To change the case sensitivity for a directory on a drive mounted to a WSL distribution (ie. Ubuntu), follow the same steps as listed above for the Windows file system.
+
+To enable case-sensitivity on a directory (FOO ≠ foo), use the command:
+
+```bash
+sudo fsutil.exe file setCaseSensitiveInfo <path> enable
```
benhillis, line 129: shouldn't need sudo here, in fact I don't think it will work.

WSL/case-sensitivity.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+To disable case-sensitivity on a directory and return to the case-insensitive default (FOO = foo), use the command:
+
+```bash
+sudo fsutil.exe file setCaseSensitiveInfo <path> disable
```
benhillis, line 135: same here

WSL/case-sensitivity.md
mattwojo: Oops! Good catch - thanks!

benhillis (approved):

PR #1263, opened 2021-08-24
Remove ~ path segment from command to run Mongo instance

16sheep:
We were creating `data/db` in `/home` but then tried to run the instance with path `~/data/db` which failed as
`home/accountname/data/db`  did not exist 

```
9. Make a directory to store data: `mkdir -p data/db`
10. Run a Mongo instance: `sudo mongod --dbpath ~/data/db`
```
This pr changes the command to run instance with home path.


craigloewen-msft:
Merged this in, thank you for filing it!

PR #1276, opened 2021-09-07
Fix typo in wsl-database

Sejdens:


mattwojo:
Great catch! Thanks @Sejdens!!

PR #1284, opened 2021-09-20
Document Git Configuration for Azure Repos

reynoldsbd:
Per GCM Core docs, some additional configuration is needed in order to
authenticate with Azure Repos.

https://github.com/microsoft/Git-Credential-Manager-Core/blob/main/docs/configuration.md#credentialusehttppath


craigloewen-msft:
Merged this! Thank you for the change!

Issue #1054, opened 2021-01-08
wsl.exe --install not running


ylr-research:
The first instruction is to run wsl --install and instead of actually running that command it says that command doesn't exist:

```
Invalid command line option: --install
Copyright (c) Microsoft Corporation. All rights reserved.

Usage: wsl.exe [Argument] [Options...] [CommandLine]

Arguments for running Linux binaries:

    If no command line is provided, wsl.exe launches the default shell.

    --exec, -e <CommandLine>
        Execute the specified command without using the default Linux shell.

    --
        Pass the remaining command line as is.

Options:
    --distribution, -d <Distro>
        Run the specified distribution.

    --user, -u <UserName>
        Run as the specified user.

Arguments for managing Windows Subsystem for Linux:

    --export <Distro> <FileName>
        Exports the distribution to a tar file.
        The filename can be - for standard output.

    --import <Distro> <InstallLocation> <FileName> [Options]
        Imports the specified tar file as a new distribution.
        The filename can be - for standard input.

        Options:
            --version <Version>
                Specifies the version to use for the new distribution.

    --list, -l [Options]
        Lists distributions.

        Options:
            --all
                List all distributions, including distributions that are currently
                being installed or uninstalled.

            --running
                List only distributions that are currently running.

            --quiet, -q
                Only show distribution names.

            --verbose, -v
                Show detailed information about all distributions.

    --set-default, -s <Distro>
        Sets the distribution as the default.

    --set-default-version <Version>
        Changes the default install version for new distributions.

    --set-version <Distro> <Version>
        Changes the version of the specified distribution.

    --shutdown
        Immediately terminates all running distributions and the WSL 2 lightweight utility virtual machine.

    --terminate, -t <Distro>
        Terminates the specified distribution.

    --unregister <Distro>
        Unregisters the distribution.

    --help
        Display usage information.
```
here's my build number

![image](https://user-images.githubusercontent.com/63487874/104027230-150bbd80-5195-11eb-9a4f-6c1f999619f8.png)

ylr-research:
I got it to work. the correct way is `wsl.exe --install -d Ubuntu`.

rhy-ama:
Doesnt work (``--install`` or ``--install -d Ubuntu``) on 10 N 1909 build 18363.1440

mbehnasr:
i have same issue

alistair0adams:
[...]

PR #1288, opened 2021-09-24
Update install.md

samuelsaumanchan:
The current instructions to run "wsl --install" doesn't work. You have to specify a specific Linux distribution. Please update the docs. Reference this Git issue here: https://github.com/MicrosoftDocs/WSL/issues/1054


craigloewen-msft:
This is a good suggestion, this only happens when WSL is already installed. 
We're investigating fixing it, but we'd still like the default guidance to be `wsl --install` which will work as expected when WSL is not installed

PR #1293, opened 2021-09-27
Fix a typo in curl in command

matifali:


craigloewen-msft:
Merged! Thank you for the fix!

PR #1359, opened 2021-11-19
fix Case-insensitive description

Joerg-Schoemer:
- Case-insenstive -> Case-insensitive
- FOO.text -> FOO.txt


Joerg-Schoemer:
There is still a typo in 
~~case-insenstive~~
**case-insensitive**

mattwojo:
Thanks @Joerg-Schoemer!

PR #1401, opened 2021-12-23
Bulk fix - Removing .localizationpriority

v-alje:
This is a bulk fix - According to the new DevRel GX group mission, the Localization production team drives the localization priority and translation quality per customer usage data, e.g. Page View or MAU of the language. The localization priority is defined by the project setting in localization pipeline - ms.localizationpriority in the source file is no longer used as a driving parameter to define the localization priority. https://review.docs.microsoft.com/help/contribute/localization-checklist-for-writer?branch=master#stop-using-mslocalizationpriority-metadata


v-alje:
#sign-off

PR #1406, opened 2021-12-29
Update wsl2-mount-disk.md

GarethTheNerd:
Changed type of command to bash. The command isn't a Powershell command


mattwojo:
Good catch! Thanks @GarethTheNerd! Appreciate the fix!

PR #1414, opened 2022-01-08
fix: Broken Markdown in breadcrumb navigation

bennycode:
My fix repairs the currently broken Markdown on https://docs.microsoft.com/en-us/windows/wsl/install:

image


mattwojo:
Good catch. Thanks @bennycode

PR #1426, opened 2022-01-20
WSL/use-custom-distro.md: Remove redirect from wheel command note.

vvavrychuk:
This redirect is actually not present in the commands block to which this note is related.


craigloewen-msft:
Great change! Thank you for the contribution.

Issue #1422, opened 2022-01-14
wsl --install doesn't work


ramanova:
The first command from the document doesn't work:
wsl --install

Getting:
`Copyright (c) Microsoft Corporation. All rights reserved.

Usage: wsl.exe [Argument] [Options...] [CommandLine]

Arguments for running Linux binaries:

    If no command line is provided, wsl.exe launches the default shell.

    --exec, -e <CommandLine>
        Execute the specified command without using the default Linux shell.

    --
        Pass the remaining command line as is.

Options:`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0cc9a77f-e6a1-a719-c337-7ff7506a1ac8
* Version Independent ID: 7ed8fb81-5b8f-4d56-7b89-bf600ce6080a
* Content: [Install WSL](https://docs.microsoft.com/en-us/windows/wsl/install)
* Content Source: [WSL/install.md](https://github.com/MicrosoftDocs/wsl/blob/main/WSL/install.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

crramirez:
https://github.com/MicrosoftDocs/WSL/issues/1410#issuecomment-1005894224

craigloewen-msft:
Closing this , as @crramirez pointed out please see the comments he provided for an answer here!

Issue #1410, opened 2022-01-05
After upgrade to Windows 11, `wsl --install` command is not enough. Asks for additional parameters.


arnisjuraga:
In documentation is said, that latest Windows versions can install WSL2 by running simple `wsl --install` 

When I did this on Windows 11 (After upgrade from Windows 10 with installed wsl already), command is not working. 
Output:

``` 
C:\Users\Arnis>wsl --install
Copyright (c) Microsoft Corporation. All rights reserved.

Usage: wsl.exe [Argument] [Options...] [CommandLine]

Arguments for running Linux binaries:

    If no command line is provided, wsl.exe launches the default shell.

    --exec, -e <CommandLine>
        Execute the specified command without using the default Linux shell.

    --
        Pass the remaining command line as is.

...
``` 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0cc9a77f-e6a1-a719-c337-7ff7506a1ac8
* Version Independent ID: 7ed8fb81-5b8f-4d56-7b89-bf600ce6080a
* Content: [Install WSL](https://docs.microsoft.com/en-us/windows/wsl/install)
* Content Source: [WSL/install.md](https://github.com/MicrosoftDocs/wsl/blob/main/WSL/install.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

crramirez:
Hello,

This command works for installing wsl. As you said in the description you have wsl already installed. This is why is not working. Uninstall wsl and install it again with wsl --install

Regards

arnisjuraga:
To be honest, I don't want to "uninstall" it. 
I have upgraded Windows and I want to be able to continue to use the same WSL2 distro I have been using before.
Is there safe way to do that, by keeping my old distro filesystem and configuration at least?

craigloewen-msft:
You can use `wsl --install -d <DistroName>` to install further distros, since the base layers of WSL are already installed. We've added better output for this in the latest WSL version. 

Please see the commands here:
https://docs.microsoft.com/en-us/windows/wsl/install#change-the-default-linux-distribution-installed

And you can run `wsl -l -v` to get a list of what distros you already have installed.

Please try running `wsl.exe` to go into a WSL instance. 

Thanks for filing!

crramirez:
[...]

Issue #1408, opened 2021-12-31
wsl --install doesn't work


SamZhangQingChuan:
```
wsl --install
```

can not run successfully.

It gives
```
❯ wsl --install
Copyright (c) Microsoft Corporation. All rights reserved.

Usage: wsl.exe [Argument] [Options...] [CommandLine]

Arguments for running Linux binaries:

    If no command line is provided, wsl.exe launches the default shell.

    --exec, -e <CommandLine>
        Execute the specified command without using the default Linux shell.

    --
        Pass the remaining command line as is.

Options:
    --cd <Directory>
        Sets the specified directory as the current working directory.
        If ~ is used the Linux user's home path will be used. If the path begins
        with a / character, it will be interpreted as an absolute Linux path.
        Otherwise, the value must be an absolute Windows path.

    --distribution, -d <Distro>
        Run the specified distribution.

    --user, -u <UserName>
        Run as the specified user.

    --system
        Launches a shell for the system distribution.

Arguments for managing Windows Subsystem for Linux:

    --help
        Display usage information.

    --install [Options]
        Install additional Windows Subsystem for Linux distributions.
        For a list of valid distributions, use 'wsl --list --online'.

        Options:
            --distribution, -d [Argument]
                Downloads and installs a distribution by name.

                Arguments:
                    A valid distribution name (not case sensitive).

                Examples:
                    wsl --install -d Ubuntu
                    wsl --install --distribution Debian

    --set-default-version <Version>
        Changes the default install version for new distributions.

    --shutdown
        Immediately terminates all running distributions and the WSL 2
        lightweight utility virtual machine.

    --status
        Show the status of Windows Subsystem for Linux.

    --update [Options]
        If no options are specified, the WSL 2 kernel will be updated
        to the latest version.

        Options:
            --rollback
                Revert to the previous version of the WSL 2 kernel.

Arguments for managing distributions in Windows Subsystem for Linux:

    --export <Distro> <FileName>
        Exports the distribution to a tar file.
        The filename can be - for standard output.

    --import <Distro> <InstallLocation> <FileName> [Options]
        Imports the specified tar file as a new distribution.
        The filename can be - for standard input.

        Options:
            --version <Version>
                Specifies the version to use for the new distribution.

    --list, -l [Options]
        Lists distributions.

        Options:
            --all
                List all distributions, including distributions that are
                currently being installed or uninstalled.

            --running
                List only distributions that are currently running.

            --quiet, -q
[...]

PR #1433, opened 2022-01-24
Add a note in case that WSL is already installed #1422 #1410 #1408 #1054

crramirez:
Looks like many people need this to be clarified. Even though a better message appear in the latest WSL version from store, it won't arrive soon to Windows 10 users


crramirez:
Hello @craigloewen-msft,

Please check this PR to see if the wsl --install problem issues decreases.

Regards,

craigloewen-msft:
Merged! Thank you for the much needed addition here @crramirez

Issue #1301, opened 2021-10-07
wsl.conf: Documnent ldconfig option


mook-as:
https://github.com/MicrosoftDocs/WSL/blob/73e7582da51ece4e975022c28c471e0bcf181110/WSL/wsl-config.md#automount

There appears to be an undocumented `ldconfig` option that defaults to `true` (in the `[automount]` section).
When set, `init` appears to drop a file in `/etc/ld.so.conf.d` called `ld.wsl.conf` which adds `/usr/lib/wsl/lib` to the search path, then runs `/sbin/ldconfig`.  That file also mentions that `[automount]` `ldconfig = false` can be used to disable this behaviour.

Presumably this is similar to the `mountFsTab` option.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: deab370f-a7a1-ed3f-89ae-22194fd8fb78
* Version Independent ID: 9f406c77-415c-80db-d596-0a7fa4ea2ba9
* Content: [Configure Linux distributions](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#ways-to-run-wsl)
* Content Source: [WSL/wsl-config.md](https://github.com/MicrosoftDocs/wsl/blob/master/WSL/wsl-config.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

Agrover112:
@craigloewen-msft  Any update on this?

craigloewen-msft:
This is a dev feature that isn't intended for public use, so we'll keep it undocumented but available. Thanks for the ping here!

Issue #1171, opened 2021-04-28
Step 3 instructions


ctaylor-ntirety:
[Step Three **Restart**   needs to be **BOLDER LARGER AND HIGHLIGHTED !** ]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0cc9a77f-e6a1-a719-c337-7ff7506a1ac8
* Version Independent ID: 7ed8fb81-5b8f-4d56-7b89-bf600ce6080a
* Content: [Install WSL on Windows 10](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
* Content Source: [WSL/install-win10.md](https://github.com/MicrosoftDocs/wsl/blob/master/WSL/install-win10.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

frank-dutmers:
I agree.

craigloewen-msft:
This is clearer in the latest and we added an inproduct ask for restart as well. Thanks for filing this!

PR #1454, opened 2022-02-10
Adding ldconfig option to automount options

Agrover112:
#1301  Added the ldconfig option (but I think you could improve the description @craigloewen-msft if it is accepted)
#1171  Just made restarting --> bold for readability.


craigloewen-msft:
Hi @Agrover112 this feature is actually undocumented as it wasn't intended to be released for public use, for now we'll keep it this way so I will close this PR. Thank you for finding this and your suggestion!

Issue #1462, opened 2022-02-20
Clarify architecture compatibility for USB device support in WSL


miitchyy:
Hi team,

I think it should be called out clearly in the prerequisites section of this guide that the `usbipd-win` dependency does not support the `win-arm64` architecture.

I did just pick up a Surface Pro X with the expectation this would work before realising that due to reasons beyond the project author's control this isn't possible (dorssel/usbipd-win#233).

I understand this feature is still in its infancy but would love to see this platform on the roadmap in the future.

Kind regards,
Mitchell O'Sullivan

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e99de78d-d27a-664d-98ba-91116b7157c8
* Version Independent ID: e99de78d-d27a-664d-98ba-91116b7157c8
* Content: [Connect USB devices](https://docs.microsoft.com/en-us/windows/wsl/connect-usb)
* Content Source: [WSL/connect-usb.md](https://github.com/MicrosoftDocs/wsl/blob/main/WSL/connect-usb.md)
* Product: **dev-environment**
* Technology: **windows-subsystem-for-linux**
* GitHub Login: @craigloewen-msft
* Microsoft Alias: **crloewen**

mattwojo:
Thanks for filing this @miitchyy -- I've just added a note to the prereqs. Wasn't aware of the Arm64 support issue until now... bummer! I'll make sure that the right folks are aware of the issue.

PR #1476, opened 2022-03-04
Add prereq note about arm support

mattwojo:


mattwojo:
@benmcmorran - Could you confirm this prereq is accurate?
Addresses https://github.com/MicrosoftDocs/WSL/issues/1462

benmcmorran:
I'm not sure why there would be an ARM64 requirement. @dorssel do you expect usbipd-win to work with ARM64?

benmcmorran:
Never mind, I see now that the dependency is coming from the drivers. This looks good to me.

benmcmorran (approved):

PR #1490, opened 2022-03-23
Added note for SQL Server support on WSL

rwestMSFT:
SQL Server on Linux requires **systemd**, which is not available on WSL.

Wording has been confirmed by @amvin87.


mattwojo:
Thanks @rwestMSFT! Appreciate the clarification here.

PR #1496, opened 2022-03-31
Correct wsl.config to wsl.conf

Zanthed:
wsl.config doesn't exist and is meant to be wsl.conf  here

Signed-off-by: June <june@eridan.me>


mattwojo:
Good catch @Zanthed - thank you! I was mixing up wsl.conf and .wslconfig. Appreciate the correction!!!

PR #1516, opened 2022-05-12
replace apt-get with apt

Raton-Laveur:
apt-get no longer supported on wsl debian


pmartincic:
I just grabbed debian via `wsl --install debian` and had both `apt` and `apt-get` present.

PR #1525, opened 2022-05-21
Fix `git-credential-manager-core.exe` path

oliver-leung:
The path to the `git-credential-manager-core.exe` has now changed to be under `bin/`, breaking references to it from WSL.


pmartincic:
Addressed as of https://github.com/MicrosoftDocs/WSL/commit/c4597be02822b80d77282f364568a7c87a9b836c

PR #1548, opened 2022-06-11
Update wsl-git.md

adiazcan:
With the latest version of Git for Windows, the installation folder has a different path.


benhillis:
sorry already merged another version of the fix, thanks for posting!

PR #1571, opened 2022-07-12
Make linux-tools-generic install instructions version agnostic

Seldaek:


Seldaek:
Well I only noticed this didn't work for me as is with WSL2 on latest kernel and ubuntu22.04. So I figured this way it is more likely to work for more people, but it may still not work in all cases in the future I suppose.. Still seems like an improvement to me but I'll leave it up to you to decide :)

mattwojo:
@benmcmorran - Any thoughts on this?

craigloewen-msft:
Great suggestion! I've merged and added this.

PR #1589, opened 2022-08-18
Updated USBIP command

SaadAnsari17:
Updated USBIP command, the previous command is not working on ubuntu 22.04


craigloewen-msft:
This feedback was addressed in this commit: https://github.com/MicrosoftDocs/WSL/commit/af971f50b76b043909caff2170e1cf19a75c46bd

Thanks for the feedback!

PR #1594, opened 2022-08-19
Networking patch

craigloewen-msft:
Added a networking section to troubleshooting


WSL/enterprise.md
```diff
@@ -38,6 +38,12 @@ Using Linux configuration manager tools is strongly recommended for monitoring a
 
 Currently, WSL offers limited control mechanisms in regard to modifying the user experience in an Enterprise scenario. Enterprise features continue in development however, below are the areas of supported and unsupported features. To request a new feature not covered in this list, file an issue in our [GitHub repo](https://github.com/microsoft/WSL/issues?q=is%3Aissue+is%3Aopen+enterprise).
 
+### Configuring WSL firewall rules
+
+If the [local policy merge](https://docs.microsoft.com/openspecs/windows_protocols/ms-gpfas/2c979624-900a-4b6e-b4ef-09b387cd62ab) firewall rule ([more details here](https://docs.microsoft.com/windows/security/threat-protection/windows-firewall/best-practices-configuring#establish-local-policy-merge-and-application-rules)) is set to "No" then WSL networking will not work. You can fix this by adding this configuration to Windows firewall settings:
```
matthewige, line 43: Suggest: If the local policy merge firewall **setting**.

This is a firewall setting, not a rule.

matthewige: Similarly, the last line should say 'You can fix this by adding this **firewall rule** to the Windows Firewall settings'

mattwojo:
I'd like to take a crack at updating the screenshot before merging this.

mattwojo (approved):
I replaced the screenshot and made a few minor editorial updates. (I would avoid using the term "fixed" as some contexts may have this set intentionally..?)

Biswa96:
Just curious. Wouldn't it be easier to add firewall rule if New-NetFirewallRule or netsh command was added instead of using wf.msc?

matthewige (approved):

PR #1618, opened 2022-09-21
Doc update request

muhyilmaz:
Windows Subsystem for Linux (WSL) updated for see better. I cannot see what is required for docker setup


craigloewen-msft:
We'll keep this formatting to maintain our current style guides. Thank you for the feedback though!

PR #1627, opened 2022-09-27
Fixed wslconfig value types

cybernop:
From my understanding `kernel` and `swapfile` in `.wslconfig` should be of type `path` as they represent Windows paths and are also handled in the example like it is described for the value type `path`.

### DIFF
--- a/WSL/wsl-config.md
+++ b/WSL/wsl-config.md
@@ -208,13 +208,13 @@ Section label: `[wsl2]`
 
 | key | value | default | notes|
 |:----|:----|:----|:----|
-| kernel | string | The Microsoft built kernel provided inbox | An absolute Windows path to a custom Linux kernel. |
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
At first glance, I was thinking that maybe the WSL team doesn't consider there to be a difference between a "string" and a "path", but then I noticed the important part of the doc immediately below that table (that you reference, but don't quote):

> Entries with the `path` value must be Windows paths with escaped backslashes, e.g: `C:\\Temp\\myCustomKernel`

So yes, I absolutely agree that the table should be updated (and was probably originally meant) to have those specified as `path`.  Good catch, IMHO!

benhillis (approved):

craigloewen-msft:
Merged! Thank you for the contribution!

PR #1639, opened 2022-10-18
Add documentation about safeMode

crramirez:
Please add **hacktoberfest-accepted** to the PR if you accept it.

### DIFF
--- a/WSL/wsl-config.md
+++ b/WSL/wsl-config.md
@@ -213,7 +213,7 @@ Section label: `[wsl2]`
 | processors | number | The same number of processors on Windows | How many processors to assign to the WSL 2 VM. |
 | localhostForwarding | boolean | `true` | Boolean specifying if ports bound to wildcard or localhost in the WSL 2 VM should be connectable from the host via `localhost:port`. |
 | kernelCommandLine | string | Blank | Additional kernel command line arguments. |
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
Recommend adding:

> Available only in Windows 11 with WSL release 0.66.2+.

This matches the terminology used on the Systemd text earlier in the doc.

crramirez:
Ready I added the requirements in the Notes field

NotTheDr01ds:
I hate to be picky (but I am), and note that I'm not on the WSL team (just someone looking through the PRs), but the new text is a bit "wordy" for the notes.  Look at the other example "notes" (e.g. `guiApplications`) where they keep this short-and-sweet:

> Only available for Windows 11.

In this case, it's not just Windows 11, but also the WSL release, which is why I recommended:

> Available only in Windows 11 with WSL release 0.66.2+.

But the MS team may want to trim that down even more.  And now I'm noticing that I reversed the MS wording of "Available only" vs. "Only available", but that's my own personal nitpick with myself ;-).

Also a heads-up that, from experience, PRs aren't reviewed all that frequently by the WSL team.  Someone does get to it though.  Just want to advise "patience" ;-).

crramirez:
Don't worry, I hope after your filter that the team won't have problems merging it.

craigloewen-msft:
Made a small edit, but overall great change, thank you for the contribution!

PR #1644, opened 2022-11-01
Fix Windows build number introducing WSL

iamCristYe:
WSL was initially introduced in 2016, while WSL 2 was introduced in 2019.
ref: https://www.howtogeek.com/248177/whats-new-in-windows-10s-anniversary-update/#:~:text=Windows%20Gets%20Its%20Own%20Linux%20Command%20Line


craigloewen-msft:
Great spot! Thank you for the contribution!

PR #1653, opened 2022-11-16
Store commands update

craigloewen-msft:
Updated the commands to include the latest for the Microsoft Store version of WSL

### DIFF
--- a/WSL/basic-commands.md
+++ b/WSL/basic-commands.md
@@ -7,7 +7,7 @@ ms.topic: article
 
 # Basic commands for WSL
 
-The WSL commands below are listed in a format supported by PowerShell or Windows Command Prompt. To run these commands from a Bash / Linux distribution command line, you must replace `wsl` with `wsl.exe`. For a full list of commands please run `wsl --help`. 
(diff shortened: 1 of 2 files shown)


WSL/basic-commands.md
```diff
@@ -1,29 +1,33 @@
...
 
-## Install a specific Linux distribution
+Options include: 
 
-```powershell
-wsl --install --distribution <Distribution Name>
-```
+- `--distribution`: Specify the Linux distribution to install. You can find available distributions by running `wsl --list --online`. 
+- `--no-launch`: Install the Linux distribution but do not launch it automatically.
+- `--web-download`: Install from GitHub rather than using the Microsoft Store. *(This will result in not receiving as-available updates from the store.)*
```
OneBlue, line 24: Is that actually the case ? Not sure if the store might not be able of recognizing the package even if installed via GH

OneBlue (approved):

mattwojo:
@craigloewen-msft -- Made a few updates, please take a look. Otherwise looks good to me.

WSL/basic-commands.md
craigloewen-msft, line 24:
```suggestion
- `--web-download`: Install from an online source rather than using the Microsoft Store.
```

PR #1659, opened 2022-11-25
Update command to perfect apply

habolanos:


craigloewen-msft:
Great spot, thank you for the contribution!

PR #1661, opened 2022-11-26
Delete duplicate manual installation instructions, fixes #1660

kanyeceejayA:
Deleted a duplicate instruction explaining how to get step-by-step instructions for instructions. Fixes Issue #1660

PR #1664, opened 2022-11-27
Delete duplicate sentence in install.md

mjtorr:


mjtorr:
Oh, never mind, already done in #1661.

PR #1667, opened 2022-11-30
Info about support for --mount in Store version

kyanha:
The Windows Store version of WSL supports --mount, even on Windows 10.

### DIFF
--- a/WSL/wsl2-mount-disk.md
+++ b/WSL/wsl2-mount-disk.md
@@ -17,7 +17,7 @@ If you are looking for guidance on how to connect a USB device (flash drive, SD 
 
 ## Prerequisites
 
-You will need to be on Windows 11 Build 22000 or later, or be running the Microsoft Store version of Windows Services for Linux, to access this feature. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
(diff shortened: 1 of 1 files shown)


WSL/wsl2-mount-disk.md
```diff
@@ -17,7 +17,7 @@ If you are looking for guidance on how to connect a USB device (flash drive, SD
 
 ## Prerequisites
 
-You will need to be on Windows 11 Build 22000 or later to access this feature. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
+You will need to be on Windows 11 Build 22000 or later, or be running the Microsoft Store version of Windows Services for Linux, to access this feature. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
```
phil-blain, line 20: WSL is for "Windows Subsystem for Linux". In any case I think just using the abbreviation would be fine:

```suggestion
You will need to be on Windows 11 Build 22000 or later, or be running the Microsoft Store version of WSL. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
```

kyanha:
I do understand that WSL is Windows Subsystem for Linux. The Store version, however, doesn't actually use the WSL abbreviation, and I had been thinking that it could be reminded. Not going to fight on it, though.

craigloewen-msft:
Awesome contribution, thank you for the feedback!

PR #1668, opened 2022-11-30
Added Microsoft Store version of WSL and prereqs.

kyanha:
KB5020030 will be released to the Release track on Dec 2022 Patch Tuesday. It is a prerequisite for at least 21H2 to run the Windows Store version of WSL.


craigloewen-msft:
This isn't necessarily a pre-req, we already have a place for WSL Store info at: https://aka.ms/wslstoreinfo

So I'll close this out. Thank you for the contribution!

PR #1670, opened 2022-12-01
Remove duplicated paragraph

ushuz:


ushuz:
It's got fixed in #1700

PR #1674, opened 2022-12-06
Update gpu-compute.md

innns:
You may need to use `sudo service docker start` to start docker.


craigloewen-msft:
Great catch! Thank you for the contribution.

PR #1678, opened 2022-12-07
Add a page on how to repair a WSL2 disk

OneBlue:
This change adds a new tutorial on how to recover a WSL2 disk in a bad state (either because it's full or because repair is needed)


craigloewen-msft:
Added @mattwojo as a reviewer here as well!

The overall content looks good to me, but I don't think that this warrants being its own top level page since the content is more niche. 

Could we add this to the "Troubleshooting" Page? Or perhaps rename the "Expand virtual hard disk size" page to "Manage your virtual hard disk" and add this as a section to that page?

OneBlue:
Thank you for the suggestion @craigloewen-msft. Merged both pages into one 'Disk maintenance' page that cover both expanding and repairing VHD's

craigloewen-msft:
@mattwojo can you help merge in this PR? The file name changed so we would likely need redirection help here

mattwojo:
@craigloewen-msft -- To confirm, we want this new "Disk maintenance operations on WSL2 distributions" page to _replace_ the "Expand the size of your WSL 2 Virtual Hard Disk" page, yes?

 @OneBlue - Thanks for contributing this! It looks great -- really appreciate the additional clarity it provides!

mattwojo:
I'm doing a bit more digging and our "Expand the size of your WSL 2 Virtual Hard Disk" page gets a lot of traffic, so I'm hesitant to  merge/redirect it and thinking that the original recommendation to create a separate "Recover / Repair" doc is the better direction to take. I'm also doing a bit of SEO research to see what folks are searching for that may be related... the only keyword queries that are getting much traffic are: "optimize vhd" and "wsl disk space" ...not much in regard to repairing, recovering, restoring... but I'll come up with a recommendation by EOD.

PR #1687, opened 2022-12-16
New how tos: locate vhd, display disk space, repair

mattwojo:


WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+sdc    8:32   0   1.5T  0 disk
+sdd    8:48   0     1T  0 disk /mnt/wslg/distro
+```
+
+Information about the block device includes:
+
+- **NAME**: The name assigned to the device will be sd[a-z], referring to the SCSI Disk with a letter designation for each disk being used. `sda` is always the system distribution.
+- **MAJ:MIN**: Represents numbers used by the Linux kernel to internally identify the devices with the first number representing the device type (8 is used for Small Computer System Interface/SCSI disks, an alternative to USB).
+- **RM**: Let's us know if the device is removable (1) or not (0).
+- **SIZE**: lets us know how much space the device is using.
```
OneBlue, line 150: SIZE is the total size of the volume, not the amount of space in use

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+NAME MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
+sda    8:0    0 363.1M  1 disk
+sdb    8:16   0     8G  0 disk [SWAP]
+sdc    8:32   0   1.5T  0 disk
+sdd    8:48   0     1T  0 disk /mnt/wslg/distro
+```
+
+Information about the block device includes:
+
+- **NAME**: The name assigned to the device will be sd[a-z], referring to the SCSI Disk with a letter designation for each disk being used. `sda` is always the system distribution.
```
OneBlue, line 147: This section feels a bit like too much information for the purpose of this page (and might create extra confusion, especially if we talk about 'removable' or 'USB' devices here. I'd recommend just keeping '**NAME**' here

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+- **Use%**: Percentage of disk space remaining (Used / Allocated size)
+- **Mounted on**: Directory path where the disk is mounted
+
+If you see that you are near to reaching the available amount of disk space allocated to your VHD, or have already received an error due to no disk space remaining, see the next section for steps on how to expand the maximum amount of disk space allocated to the VHD associated with your Linux distribution.
+
+> [!NOTE]
+> The size of the VHD represented by the output of the df command represents the size allocated for the Linux file system and will be different than the total size on your device used by the Linux distribution. To view the **Total usage** size of a Linux distribution installed in your `LocalApps` directory, you can use the Settings app on Windows. Go to **Settings > Apps > Installed apps** and search the list of installed apps by entering the name of the Linux distribution you want to view, for example "Ubuntu" or "Debian". By selecting `...`, you can choose **Advanced options** to view the total disk space used, amount of data stored, and version number of your installed distribution. You can also **Uninstall** the distribution.
```
OneBlue, line 44: This is necessarily true since the distribution VHDX isn't guaranteed to be stored inside the application's LocalState folder. I'd recommend removing this note

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+
+## How to expand the size of your WSL 2 Virtual Hard Disk
+
+To expand the VHD size for a Linux distribution beyond the 1TB maximum amount of allocated disk space (or beyond the 256GB maximum allocated prior to WSL release 0.58.0), follow these steps:
+
+1. Terminate all WSL instances using the command: `wsl --shutdown`
+
+2. Find your distribution installation package name ('PackageFamilyName') by opening PowerShell and entering the following command (replacing `<distribution-name>` with the name of the distribution):
+    - `Get-AppxPackage -Name "*<distribution-name>*" | Select PackageFamilyName`
```
OneBlue, line 53: Let's point to the `## How to locate the .vhdx file and disk path for your Linux distribution` section instead. This has the benefit of:

- Being simpler (since it's a one liner)
- Supporting distributions outside of their default installation path
- Supporting distributions installed outside of apps

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+- **MAJ:MIN**: Represents numbers used by the Linux kernel to internally identify the devices with the first number representing the device type (8 is used for Small Computer System Interface/SCSI disks, an alternative to USB).
+- **RM**: Let's us know if the device is removable (1) or not (0).
+- **SIZE**: lets us know how much space the device is using.
+- **RO**: Let's us know if the device is read-only (1) or not (0).
+- **TYPE**: Refers to the device type (disk in this case, could also be something like lvm for a logical volume device).
+- **MOUNTPOINTS**: Refers to the current directory on the files system where the block device is located (SWAP is for preconfigured inactive memory so no mountpoint).
+
+### Read-only fallback error
+
+If WSL encounters an error while installing (mounting) a Linux distribution, it may mount the distribution as read-only (RO = 1) as a fallback. If that happens, the distribution may display the following error during startup:
```
OneBlue, line 157: Let's remove `installing` here distribution installation failures are unrelated to this page

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+When a distribution is started as read-only, any attempts to write to the filesystem will fail with an error like this:
+
+```bash
+$ touch file
+touch: cannot touch 'file': Read-only file system
+```
+
+To repair a disk mount error in WSL, and restore it back to a usable / writeable state again, you can use the `wsl --mount` command to re-mount the disk with the following steps:
```
OneBlue, line 170: Let's use `wsl.exe` everywhere as the original pull request did (this protects from powershell aliases / potential conflicting commands)

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+    ```powershell
+    wsl.exe --unmount
+    ```
+
+> [!WARNING]
+> You can use the command: `sudo mount -o remount,rw /` to return a read-only distribution to a usable/writable state, but all changes will be in-memory and so will be lost when the distribution is restarted. We recommend using the steps listed above to mount and repair the disk instead.
+
+## How to locate the .vhdx file and disk path for your Linux distribution
+
+If you are having trouble locating the *.vhdx* file for a Linux distribution, go to the directory path `%LocalAppData%\Packages\`. Each Linux distribution that you have installed includes a subfolder specific to that distribution.
```
OneBlue, line 201: That's not true since this path can be changed

OneBlue: Let's remove that part to avoid confusion

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+
+For the default Ubuntu distribution, this directory will look something like `CanonicalGroupLimited.UbuntuonWindows_79rhkp1fndgsc`. Inside that directory will be a `LocalState` folder with the `ext4.vhdx` file that stores your data.
+
+The VHD used by WSL 2 utilizes a ext4 format for the file system, not NTFS, so Windows will not natively read this format outside of the WSL environment. (Tools like Disk Manager or diskpart installed on Windows NTFS file system will not work.)
+
+If you are having trouble locating the disk path where the *.vhdx* file is stored for a particular Linux distribution, you can use PowerShell to help you find it. Replace `<distribution-name>` with the actual distribution name and run the following command:
```
OneBlue, line 207: Let's use the powershell command as the default way of locating disks. Asking users to find applications that "look like" their distribution is a bit lose and might lead to incorrect disks if multiple versions of the same distributions are installed. The below powershell command is guaranteed to be giving the correct result

WSL/disk-space.md
```diff
@@ -0,0 +1,217 @@
...
+- [How to expand the size of the VHD](#how-to-expand-the-size-of-your-wsl-2-virtual-hard-disk)
+- [How to repair the VHD if an error occurs](#how-to-repair-a-vhd-mounting-error)
+- [How to locate the .vhdx file and disk path for any installed Linux distributions](#how-to-locate-the-vhdx-file-and-disk-path-for-your-linux-distribution)
+
+Windows Subsystem for Linux (WSL 2) uses a virtualization platform to install Linux distributions alongside the host Windows operating system, creating a Virtual Hard Disk (VHD) to store files for each of the Linux distributions that you install. These VHDs use the [ext4 file system type](https://opensource.com/article/17/5/introduction-ext4-filesystem) *(different from the NTFS file system used by Windows)* and are represented on your Windows hard drive as an *ext4.vhdx* file.
+
+WSL 2 automatically resizes these VHD files to meet storage needs. By default each VHD file used by WSL 2 is initially allocated a 1TB maximum amount of disk space (prior to [WSL release 0.58.0](https://github.com/microsoft/WSL/releases/tag/0.58.0) this default was set to a 256GB maximum).
```
OneBlue, line 19: The default before WSL 0.58.0 was 512GB, not 256GB

OneBlue (changes requested):

WSL/disk-space.md
mattwojo: Hmm. I found knowing these descriptions quite helpful and am frustrated when a command is run resulting in a list of information that I don't understand. I generally prefer to err on the side of explaining everything mentioned, but also am interested to better understand the potential for confusion... can you help clarify what you mean?

WSL/disk-space.md
mattwojo: Can you clarify what **Settings > Apps > Installed apps > Ubuntu > Advanced options > Total disk space** is defining? Seems worth clarifying in the docs since it is available and discoverable. Does the edit below look correct?

> The size of the VHD represented by the output of the df command represents the size allocated for the Linux file system and will be different than the total size on your device used by the Linux distribution. You can view total disk space used by the distribution by going to **Settings > Apps > Installed apps** and search the list of installed apps by entering the name of the Linux distribution you want to view, for example "Ubuntu" or "Debian". By selecting `...`, you can choose **Advanced options** to view total disk space used, amount of data stored, and version number of your installed distribution. You can also **Uninstall** the distribution.

WSL/disk-space.md
mattwojo: @OneBlue - Do you know when the switch happened from 256GB to 512GB? May be nice to include that reference as well.

WSL/disk-space.md
mattwojo: I'm a bit confused about the scenario for this then. A few lines earlier I wrote: "When WSL 2 installs a Linux distribution, it is mounting the distribution as a Virtual Hard Disk (VHD) with it's own file system." Isn't the scenario here that a WSL user is installing a new Linux distribution and encounters a mounting error? Maybe I need more clarity on the scenario in which they would encounter a mounting error... should this instead be in the "Mount a Linux disk in WSL 2" doc?

WSL/disk-space.md
mattwojo: Updated.

WSL/disk-space.md
mattwojo: Updated.

WSL/disk-space.md
mattwojo: Updated.

WSL/disk-space.md
```diff
@@ -0,0 +1,210 @@
...
+      Select vdisk file="<pathToVHD>"
+      ```
+
+5. Display the details associated with this virtual disk, including the **Virtual size**, representing the current maximum size the VHD is allocated:
+
+      ```cmd
+      detail vdisk
+      ```
+
+6. You will need to convert the **Virtual size** to megabytes. For example, if **Virtual size: 1024GB**, convert this to **1024000**. The new value you enter must be greater than this original value. For example, to double the virtual size listed above, you could enter the value: **2048000**.
```
mattwojo, line 76: Are there more parameters we should add here? I tried changing the size of my VHD from 1024000 to 1024010 and received an error. Does the new size have to meet certain criteria?

WSL/disk-space.md
OneBlue: `Settings > Apps > Installed apps > Ubuntu > Advanced options > Total disk space` measure the amount of the storage used by the MSIX application, which can, but isn't guaranteed to contain the distro VHDX

WSL/disk-space.md
OneBlue: This change happened inbox so it's going to be harder to track

WSL/disk-space.md
OneBlue: This could be a good documentation to point to for `lslblk`: https://linuxhint.com/linux-lsblk-command-tutorial-for-beginners/ 

Overall, I think we should avoid documenting what standard Linux command do and focus on WSL-specific content, since  well known Linux commands are usually well documented online.

I'm worried that users will get confused if we start talking about USB devices or LVM since this could create confusion which scenarios like [this one](https://github.com/microsoft/WSL/issues/7566) which are orthogonal to this

WSL/disk-space.md
OneBlue: What error did you get ?

WSL/disk-space.md
mattwojo: Updated

WSL/disk-space.md
mattwojo: No worries -- I added a more general statement about it. Thanks for looking.

WSL/disk-space.md
mattwojo: "DiskPart has encountered an error: The parameter is incorrect."

WSL/disk-space.md
mattwojo: You're right -- that gets too confusing. Removing this note. Will reassess if we get any feedback, but otherwise agreeing that including this may just add more confusion.

WSL/disk-space.md
mattwojo: I don't think that the average WSL user will understand what "a WSL distribution mounting it's disk" means. I don't... and would've assumed that it's related to using the wsl --mount command to mount a disk. Now that I've spent some time with this, I *think* that I'm beginning to understand that each time a user opens a Linux distro that has been installed, WSL goes through the process of "mounting" that ext4.vhd disk so that it can be used (?)... and that during that process of opening (mounting) the distro, an error might occur (due to a power failure or something?), which would result in needing to follow these Repair instructions.   
Is that correct?

mattwojo: Here's the update that I made -- 
"If WSL encounters a "mounting error" when opening a Linux distribution, the distribution may be set as read-only (RO = 1) as a fallback. If that happens..."

It leaves things a bit vague as to _why_ such a mounting error might occur... but I'm guessing that there are lots of different possible reasons (?)... but the main issue being resolved here is that this sort of mounting error message would not be related to: 1) Installing a distro, nor 2) using the wsl.exe --mount command. Correct?

WSL/disk-space.md
mattwojo: I removed reference to USB and LVM. We generally try to avoid linking out to 3rd party sites unless we're pretty confident of their ownship and longevity (sites like Ubuntu.com)... and in this case doesn't seem necessary.

WSL/disk-space.md
OneBlue: Did you enter 1024010 as megabytes ? That would be a really big size. Does 1025 works ? I think the above section is bit misleading because `1024GB` should be `1024MB`

WSL/disk-space.md
```diff
@@ -0,0 +1,207 @@
...
+- **MAJ:MIN**: Represents numbers used by the Linux kernel to internally identify the devices with the first number representing the device type (8 is used for Small Computer System Interface/SCSI disks).
+- **RM**: Let's us know if the device is removable (1) or not (0).
+- **SIZE**: Total size of the volume.
+- **RO**: Let's us know if the device is read-only (1) or not (0).
+- **TYPE**: Refers to the device type (disk in this case).
+- **MOUNTPOINTS**: Refers to the current directory on the files system where the block device is located (SWAP is for preconfigured inactive memory so no mountpoint).
+
+### Read-only fallback error
+
+If WSL encounters a "mounting error" when opening a Linux distribution, the distribution may be set as read-only (RO = 1) as a fallback. If that happens, the distribution may display the following error during startup:
```
OneBlue, line 151: I think we should remove the `(RO = 1)` because might create confusion between read-only devices (what lsblk shows) and read-only mountpoints (which is what this section is about)

WSL/disk-space.md
```diff
@@ -0,0 +1,207 @@
...
+(Get-ChildItem -Path HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss | Where-Object { $_.GetValue("DistributionName") -eq '<distribution-name>' }).GetValue("BasePath") + "\ext4.vhdx"
+```
+
+The result will display a path looking something like `%LOCALAPPDATA%\Packages\<PackageFamilyName>\LocalState\<disk>.vhdx`. For example:
+
+```powershell
+C:\Users\User\AppData\Local\Packages\CanonicalGroupLimited.UbuntuonWindows_79rhkp1fndgsc\LocalState\ext4.vhdx
+```
+
+Copy the directory path to this `ext4.vhdx` file by highlighting it and either right-clicking or selecting Ctrl+C.
```
OneBlue, line 207: Let's remove this since it might not be true depending on the terminal that the user is using and its configuration

OneBlue (changes requested):

WSL/disk-space.md
OneBlue: Having a second look at this, should we simply instruct users to use [Resize-Vhd](https://learn.microsoft.com/en-us/powershell/module/hyper-v/resize-vhd?view=windowsserver2022-ps) instead ? It's much simpler to use than diskpart in this case and we've seen users corrupt their disks with diskparts (see https://github.com/microsoft/WSL/issues/9443)

WSL/disk-space.md
mattwojo: Updated to remove the bit about highlighting/right-clicking which may be terminal-specific.

WSL/disk-space.md
mattwojo: Removed. Thanks for pointing out this differentiation!

WSL/disk-space.md
mattwojo: We can use whatever you think is best. Could you take a crack at rewriting this part? I've not used Resize-VHD yet.

mattwojo: Also, can that Resize-Vhd command be used to reduce/shrink the size of the ext4.vhdx file as well? I've expanded my kali-linux vhd to a large max, but cannot figure out how to return it to the original size. Maybe we should include a bit about this as well?

OneBlue: Just pushed a commit on this branch that does that, let me know what you think. I also made some simplification to the resize2fs part, since it can actually detect the correct size on its own

OneBlue: Unfortunately, reducing the size of the VHD is much harder operation, since the filesystem needs be shrinked first. If we decide that we need to do that, let's do it in a followup change to unblock this one

mattwojo: Added a note to be cautious not to accidentally expand the vhd beyond what you need. Maybe we just see if there are issues that come in asking about reducing vhd size before we add more content to this doc that could potentially be confusing.

PR #1692, opened 2022-12-25
Fixed messed bcdedit example

stesee:


craigloewen-msft:
This was fixed in a previous PR, thank you for the feedback here!

Issue #1694, opened 2022-12-28
Cannot set password for mysql, thus cannot connect to mysql service using username and password.


Winson-Huang:
### Documentation Issue

https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-database#install-mysql
https://github.com/MicrosoftDocs/WSL/blob/main/WSL/tutorials/wsl-database.md#install-mysql

my distro is Ubuntu, when I run the security script prompts using command `sudo mysql_secure_installation` , and set password according to the prompt, it print some error info:

```
... Failed! Error: SET PASSWORD has no significance for user 'root'@'localhost' as the authentication method used doesn't store authentication data in the MySQL server. Please consider using ALTER USER instead if you want to change authentication parameters.
```

So, I cannot set password for root and any other users of mysql, thus I cannot connect to mysql service using username and password. At present, it seems that the only way to get a mysql command prompt is `sudo mysql` command in WSL shell. 

However, most database client need username and password when config a new connection, that means the mysql service in WSL basically cannot be used in development and programming. 

### Link to documentation page

https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-database#install-mysql
https://github.com/MicrosoftDocs/WSL/blob/main/WSL/tutorials/wsl-database.md#install-mysql

### Suggested Improvements

WSL document should list a correct way to set password for MySQL.

PR #1695, opened 2022-12-28
Fix issue #1694

Winson-Huang:
In order to set password for MySQL root user, after installation of mysql-server, change the authentication plugin into mysql_native_password.


Winson-Huang:

Winson-Huang:
@mattwojo  Please review this pull request, thank you.

mattwojo:
@asklar - Can you confirm this fix?

Winson-Huang:
Two months later, you havn't even review this change...

mattwojo:
@Winson-Huang  - Thank you for this contribution to the WSL docs. We did review it, but the story was changing for WSL because of the systemd implementation. I'm not a MySQL expert, so I was trying to get help and we did a rewrite of most of this database tutorial because now that systemd is implemented in WSL, it is more in parity with any other Linux instructions. I believe that [what we have on the doc now](https://learn.microsoft.com/windows/wsl/tutorials/wsl-database#install-mysql) is more inline with this change that you proposed, so I'm closing this PR, but please take a look and if you feel that anything is incorrect or should be framed differently, please feel welcome to make an update and I'll commit to reviewing it right away. Thanks again for your contribution and my apologies that it took so long to resolve.

mattwojo:
*I am including your addition about running the security script-- thanks for including that!

PR #1697, opened 2022-12-29
Update install.md

Awadhesh43:
Duplicate description text for Install Linux on Windows with WSL page

Issue #1696, opened 2022-12-29
Duplicate description text for Install Linux on Windows with WSL page


Awadhesh43:
### Documentation Issue

The below mentioned description under heading Install WSL can be seen twice. May be an update is required to avoid confusion and for better readability purpose.

### Link to documentation page

https://learn.microsoft.com/en-us/windows/wsl/install#:~:text=be%20changed).-,If%20you%27re%20running%20an%20older%20build%2C%20or%20just%20prefer%20not%20to%20use%20the%20install%20command%20and%20would%20like%20step%2Dby%2Dstep%20directions%2C%20see%20WSL%20manual%20installation%20steps%20for%20older%20versions.,-If%20you%27re%20running

### Suggested Improvements

Commit - fa7237857e13e3c48cce4380b2425e8043878e22

phil-blain:
in fact this is a duplicate of #1660

craigloewen-msft:
Fixed here! https://github.com/MicrosoftDocs/WSL/pull/1697

Issue #1660, opened 2022-11-26
Duplicate Paragraph under "Install WSL command"


kanyeceejayA:
### Documentation Issue

Under the "Install WSL command" section, there's a paragraph that's repeated. The paragraph says:

> If you're running an older build, or just prefer not to use the install command and would like step-by-step directions, see [WSL manual installation steps for older versions](https://learn.microsoft.com/en-us/windows/wsl/install-manual).

It appears twice, one right under each other. 

This can be seen on the page linked below:

https://learn.microsoft.com/en-us/windows/wsl/install#install-wsl-command 

### Link to documentation page

https://learn.microsoft.com/en-us/windows/wsl/install#install-wsl-command

### Suggested Improvements

The second paragraph can be deleted

PR #1700, opened 2023-01-02
removing duplicate line

tobystic:


mattwojo (approved):

mattwojo:
Good catch. We don't need to say this twice in the same page.

ushuz:
there're a couple of related PRs and issues:
#1697 
#1696 
#1661 
#1660

PR #1702, opened 2023-01-09
Adding another tip that worked for many users

supuncbr:
WSL not responding and waiting forever issue resolving tip that worked for me and many other users


Biswa96:
In most of the cases, WSL does not need a full system reboot and restarting the LxssManager or WslService service is enough for that.

supuncbr:
yes, but  restarting the LxssManager or WslService service did not worked for me and I wasted few days troubleshooting this and only this worked

craigloewen-msft:
Hi @supuncbr , we try not to put as a troubleshooting step for users to go into the advanced restart options for Windows, as there are some potentially destructive actions a user could take by accident on doing so (Like reformatting their operating system etc.) So we'll close this PR, thank you for the contribution!

supuncbr:
Hi Craig,
No worries, cheers
Regards,
Supun

PR #1710, opened 2023-01-22
Don't instruct on using exploitable key storage

aknobloch:
The `apt-key add` method for storing signing authorities is deprecated, due to a vulnerability in which an illegitimate package could be signed by the wrong authority, and thus allowed to complete download despite being potentially malicious. See [here](https://wiki.debian.org/DebianRepository/UseThirdParty) for further reading.

This change updates to instead use the recommended keyrings approach, with the `[signed-by=` field to protect against the above exploit. This is the approach NVIDIA already recommends, for example [this installation guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html).

### DIFF
--- a/WSL/tutorials/gpu-compute.md
+++ b/WSL/tutorials/gpu-compute.md
@@ -39,11 +39,11 @@ This guide will show how to set up:
     ```
     
     ```bash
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
Thank you for the feedback here, I've merged it in!

PR #1715, opened 2023-01-27
Update basic-commands.md

djoy2:
Updated the install command as this required -d to be specified


craigloewen-msft:
In the latest Store version and on install, the `-d` flag is not required. We'll keep it as is to reflect what's in the latest product. Thank you for the contribution!

PR #1731, opened 2023-03-05
Update connect-usb.md correct hwdata installation command

MohabMohie:
updating with the correct command

'sudo apt install linux-tools-generic hwdata'
instead of the old one which doesn't work as per this issues thread https://github.com/microsoft/WSL/issues/7652


craigloewen-msft:
This has been fixed in an earlier PR, thank you for the contribution here!

lowlydba:
@craigloewen-msft The current live documentation is still wrong and provides commands that do not work - can you refer to where it was supposedly fixed?

PR #1743, opened 2023-03-17
add the adobe-target metadata for A/B testing

partychen:
@craigloewen-msft could you please help to merge this metadata change (adobe-target=true) to main branch and then to live? This is to enable A/B testing ability across many url base paths so we can improve Learn user experience based on data-driven decisions. This meta (adobe-target) itself will not change how your content is displayed or modify any other UI behaviors so it is safe to merge.


mattwojo:
@partychen - Could we please follow up via email (mattwoj@ms). I'd like to know more about this A/B Testing implementation. Thank you.

PR #1756, opened 2023-04-05
Fix weird codeblock in Server

Erisa:
Presumably fixes this 
image

On https://learn.microsoft.com/en-us/windows/wsl/install-on-server#install-wsl-on-previous-versions-of-windows-server


craigloewen-msft:
Great suggestion! Thank you for the fix.

Issue #1753, opened 2023-03-31
Incorrect path to WSL folders


felixtsem5:
### Documentation Issue

Under the **Trouble finding docker image storage folder** section,
I believe it's \\\\$wsl\\ instead of \\wsl$\\ (the two bullet points, with single-backslash), and \\\\wsl\\ (after **Enter:**, no dollar sign).

### Link to documentation page

https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-containers

mikaello:
Hm, on my computer™ it is another option that seems to be correct:

* `\\wsl$\docker-desktop`
* `\\wsl$\docker-desktop-data`

craigloewen-msft:
The usage of `\\wsl$\` is correct, you can also use `\\wsl.localhost\`

mikaello:
But the problem was also single slash, is that correct?

![Screenshot_20230419_082054_Chrome](https://user-images.githubusercontent.com/2505178/232984177-55c93da2-4053-4919-b8d4-6f6bb4488404.jpg)

PR #1759, opened 2023-04-13
docs: fix wsl file system notation

mikaello:
Fixes https://github.com/MicrosoftDocs/WSL/issues/1753


craigloewen-msft:
Closed this on the issue. The usage here is correct!

mikaello:
So it is correct to show single slash?

Screenshot_20230419_082054_Chrome

PR #1779, opened 2023-06-22
Issue resolutions

mattwojo:


WSL/wsl2-mount-disk.md
```diff
@@ -1,27 +1,37 @@
...
 > [!NOTE]
 > Administrator access is required to attach a disk to WSL 2.
 > The WSL 2 `mount` command does not support mounting a disk (or partitions that belong to the disk) that is currently in use. `wsl --mount` always attaches the entire disk even if only a partition is requested. You can't mount the Windows installation disk.
 
 ## Prerequisites
 
-You will need to be on Windows 11 Build 22000 or later, or be running the Microsoft Store version of WSL. You can join the [Windows Insiders Program](https://insider.windows.com/) to get the latest preview builds.
+You will need to be on Windows 11 Build 22000 or later, or be running the Microsoft Store version of WSL. To check your WSL and Windows version, use the command: `wsl.exe --version`
+
+## Mount an external disk
```
craigloewen-msft, line 22: What was the feedback to add this one? 

We mount the drives for you automatically... We could change this to networking drives maybe? 

```
## Mount an external drive seen on Windows

You can mount an external drive seen from Windows by creating a mounted directory, and then using the `drvfs` file system interop plugin. For example, if I wanted to mount `\\myNetworkShare\files\` in WSL I could do so with the command below, provided that I have created a /mnt/myNetworkShare/files folder in WSL:

sudo mount -t drvfs '\\myNetworkShare\files' /mnt/myNetworkShare/files

```


But I don't see the value of putting how to mount Windows drives when they're already automounted. Especially at the top of the doc.

mattwojo: Fixed this / changed the section to explain differences between mounting an NTFS formatted drive and accessing via WSL vs mounting a Ext4 formatted drive

PR #1797, opened 2023-08-08
duplicate mention of PowerShell

suelspahiu:
duplicate mention of PowerShell, remove completely or maybe mention Bash


craigloewen-msft:
Great catch! Thank you for the submission.

PR #1830, opened 2023-10-19
Fix systemd spelling

panekj:
If you have doubts how it is spelled, please see https://freedesktop.org/wiki/Software/systemd/ `Spelling` section


mattwojo:
Thank you!

PR #1838, opened 2023-11-06
Update `useWindowsDnsCache` explanation to avoid confusion

zfb132:
I've updated the `useWindowsDnsCache` setting description to match its literal meaning: 
When set to true, it uses Windows DNS cache; when false, it bypasses it. 

The current documentation describes the opposite, which is counterintuitive. 
This might be an oversight, and this PR corrects it to prevent any confusion.

### DIFF
--- a/WSL/wsl-config.md
+++ b/WSL/wsl-config.md
@@ -179,7 +179,7 @@ hostname = DemoHost
 generateHosts = false
 generateResolvConf = false
 
-# Set whether WSL supports interop process like launching Windows apps and adding path variables. Setting these to false will block the launch of Windows processes and block adding $PATH environment variables.
+# Set whether WSL supports interop processes like launching Windows apps and adding path variables. Setting these to false will block the launch of Windows processes and block adding $PATH environment variables.
 [interop]
 enabled = false
 appendWindowsPath = false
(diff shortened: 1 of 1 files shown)


WSL/wsl-config.md
```diff
@@ -247,15 +247,15 @@ Section label: `[experimental]`
 
 | Setting name | Value | Default | Notes |
 |:----|:----|:----|:----|
-|`useWindowsDnsCache`**| bool | false | Only applicable when `experimental.dnsTunneling` is set to true. When this option is set to True, DNS requests tunneled from Linux will bypass cached names within Windows to always put the requests on the wire. |
+|`useWindowsDnsCache`**| bool | false | Only applicable when `experimental.dnsTunneling` is set to true. When this option is set to false, DNS requests tunneled from Linux will bypass cached names within Windows to always put the requests on the wire. |
```
mattwojo, line 250: @craigloewen-msft Can you confirm?

mattwojo: Ok - confirmed this is correct and that we had this backwards. Thank you for the correction @zfb132!

mattwojo:
These updates look great. Thanks for the fixes @zfb132! 
Resolves https://github.com/MicrosoftDocs/WSL/issues/1844

PR #1839, opened 2023-11-06
fix broken link

zfb132:
This PR fixes the following suggestions  
> Link 'https://devblogs.microsoft.com/commandline/the-initial-preview-of-gui-app-support-is-now-available-for-the-windows-subsystem-for-linux' points to a page that doesn't exist. Check the path or URL and update the link.

This issue has been present for a long time, but since it was only a suggestion, it seems that no one has attempted to fix it. I've found an updated link and have made the corrections in the documentation.


mattwojo:
Thanks for fixing this link @zfb132!

PR #1858, opened 2023-11-23
Fix the command parameter of configuring Hyper-V firewall

zhaoqiu-china:


craigloewen-msft (approved):

mattwojo:
Thank you for identifying this issue @zhaoqiu-china! This was a great catch - very appreciated.

PR #1866, opened 2023-12-03
Fix broken code block and bookmark link

antonkesy:
The code block is not closed properly; therefore, the code is not highlighted as such


mattwojo:
Thanks @antonkesy !

PR #1880, opened 2023-12-20
Update gui-apps.md to simplify commits.

WesleyBryie:
I figured, if we are already referencing use of APT, and using it too, why use DPKG directly?

I simply changed the command to use APT to install google chrome, and its dependencies, rather than using 3 separate steps just for installing.


mattwojo (approved):

mattwojo:
Thanks @WesleyBryie !

WesleyBryie:
> Thanks @WesleyBryie !

Anytime!

PR #1884, opened 2023-12-25
Updated guide to fix an issue because of a wrong command

vmsteiner:
Running Dev Containers: **Open Folder in Container...** when using this guide is not / no longer correct, as most readers will already have opened the folder while connected to WSL based on the steps provided above. We need to use **Dev Containers: Reopen in Container** or it will fail to create the .devcontainer folder and devcontainer.json and give a (unfortunately not very clear) error message.

See also: https://code.visualstudio.com/docs/devcontainers/containers#_open-a-wsl-2-folder-in-a-container-on-windows
   
- Use the Dev Containers: Reopen in Container command from a folder already opened using the WSL extension.
- Select Dev Containers: Open Folder in Container... from the Command Palette (F1) and choose a WSL folder using the local \\wsl$ share (from the Windows side).


mattwojo:
Thanks @vmsteiner !

PR #1895, opened 2024-01-09
fix ambiguous wording for required win version

jantari:
Previously this part of the docs stated that Windows version 22H2 was required, but as I learned that is actually wrong:

https://github.com/microsoft/WSL/issues/10853

it is specifically Windows *11* version 22H2 that is required. Windows *10* version 22H2 does *not* work unfortunately.


mattwojo:
Thanks for adding this clarification @jantari !

PR #1901, opened 2024-01-16
Moving experimental settings to the experimental section

keith-horton:
Some networking config settings are still experimental. Moving those to the experimental section of the document so customers know how to use these settings.


WSL/wsl-config.md
```diff
@@ -249,6 +244,15 @@ These settings are opt-in previews of experimental features that we aim to make
 |:----|:----|:----|:----|
 |`autoMemoryReclaim`| string | disabled | Automatically releases cached memory after detecting idle CPU usage. Set to `gradual` for slow release, and `dropcache` for instant release of cached memory. |
 |`sparseVhd`| bool | false | When set to true, any newly created VHD will be set to sparse automatically. |
+|useWindowsDnsCache**| bool | false | Only applicable when `experimental.dnsTunneling` is set to true. When this option is set to false, DNS requests tunneled from Linux will bypass cached names within Windows to always put the requests on the wire. |
```
CatalinFetoiu, line 247: should we change "experimental.dnsTunneling" to "wsl2.dnsTunneling"? same for the lines below

OneBlue, line 247: The settings need to be between "`" (this will also prevent markdown issues with the double '*')

OneBlue, line 247: If we're documenting those as experimental, might be worth adding which version of WSL introduces those settings as well

WSL/wsl-config.md
```diff
@@ -249,6 +244,15 @@ These settings are opt-in previews of experimental features that we aim to make
...
+|bestEffortDnsParsing**| bool | false | Only applicable when `experimental.dnsTunneling` is set to true. When set to true, Windows will extract the question from the DNS request and attempt to resolve it, ignoring the unknown records. |
+|initialAutoProxyTimeout*| string | 1000 | Only applicable when `experimental.autoProxy` is set to true. Configures how long (in milliseconds) WSL will wait for retrieving HTTP proxy information when starting a WSL container. If proxy settings are resolved after this time, the WSL instance must be restarted to use the retrieved proxy settings. |
+|ignoredPorts**| string | null | Only applicable when `experimental.networkingMode` is set to `mirrored`. Specifies which ports Linux applications can bind to, even if that port is used in Windows. This enables applications to listen on a port for traffic purely within Linux, so those applications are not blocked even when that port is used for other purposes on Windows. For example, WSL will allow binding to port 53 in Linux for Docker Desktop, as it is listening only to requests from within the Linux container. Should be formatted in a comma separated list, e.g: `3000,9000,9090` |
```
OneBlue, line 250: `experimental.networkingMode` -> `wsl2.networkingMode`

OneBlue: (same for below)

OneBlue (changes requested):

WSL/wsl-config.md
craigloewen-msft: I'm confused by this, why should this be under `wsl2.dnsTunneling`? Isn't it an experimental flag and under `experimental.dnsTunneling`??

WSL/wsl-config.md
craigloewen-msft, line 250:
```suggestion
|ignoredPorts**| string | null | Only applicable when `wsl2.networkingMode` is set to `mirrored`. Specifies which ports Linux applications can bind to, even if that port is used in Windows. This enables applications to listen on a port for traffic purely within Linux, so those applications are not blocked even when that port is used for other purposes on Windows. For example, WSL will allow binding to port 53 in Linux for Docker Desktop, as it is listening only to requests from within the Linux container. Should be formatted in a comma separated list, e.g: `3000,9000,9090` |
```

WSL/wsl-config.md
```diff
@@ -249,6 +244,15 @@ These settings are opt-in previews of experimental features that we aim to make
...
+|hostAddressLoopback**| bool | false | Only applicable when `experimental.networkingMode` is set to `mirrored`. When set to True, will allow the Container to connect to the Host, or the Host to connect to the Container, by an IP address that's assigned to the Host. Note that the 127.0.0.1 loopback address can always be used - this option allows for all additionally assigned local IP addresses to be used as well. |
```
craigloewen-msft, line 251:
```suggestion
|hostAddressLoopback**| bool | false | Only applicable when `wsl2.networkingMode` is set to `mirrored`. When set to True, will allow the Container to connect to the Host, or the Host to connect to the Container, by an IP address that's assigned to the Host. Note that the 127.0.0.1 loopback address can always be used - this option allows for all additionally assigned local IP addresses to be used as well. |
```

WSL/wsl-config.md
craigloewen-msft: Versioning is done by adding a `*` or `**` and footnoting it below.

craigloewen-msft (approved):
Looks good! Just needs small fixes and then we are good to merge IMO.

WSL/wsl-config.md
mattwojo: Just added those `code ticks` for each of the new settings so that they show up as code.

mattwojo: Sounds like this is resolved.

keith-horton: dnsTunneling is not exclusive to [experimental], it can be specified directly under [wsl2]

keith-horton: and yep, fixing the text to not specify 'experimental' for dnsTunneling

WSL/wsl-config.md
keith-horton: yep. thanks!

OneBlue (approved):

PR #1910, opened 2024-01-25
Import export command clarification

DanKaplanSES:
- My original intent was to clarify that the export command exports a **snapshot** of the specified distribution. The way it is currently written doesn't say otherwise, but it could be interpreted to mean it exports a clean install without the user's modifications. 

    - In other words, it could be saying `wsl --export Ubuntu ubuntu.tar` outputs the equivalent of downloading the official "ubuntu.tar" from the internet. 
- Splitting these two was not part of my original goal, merely a step along the way to achieve it.


mattwojo:
Another great update -- thank you @DanKaplanSES!!

Issue #1907, opened 2024-01-21
WSL1/WSL2 comparison table not correct - WSL1 does support IPv6


hyanggi:
### Documentation Issue

As I have mentioned in #1806, WSL1 does support IPv6. The recent change (stating that it doesn't) is not correct. Please fix it.

### Link to documentation page

https://learn.microsoft.com/en-us/windows/wsl/compare-versions

### Suggested Improvements

Correctly mark that WSL1 supports IPv6.

PR #1931, opened 2024-03-01
Correct marking of IPv6 support for WSL1

krishnacx:
Fixes https://github.com/MicrosoftDocs/WSL/issues/1907

### DIFF
--- a/WSL/compare-versions.md
+++ b/WSL/compare-versions.md
@@ -27,7 +27,7 @@ Feature | WSL 1 | WSL 2
  Full system call compatibility| ❌ | ✅
  Performance across OS file systems| ✅ | ❌
  systemd support| ❌ | ✅
- IPv6 support| ❌ | ✅
+ IPv6 support| ✅ | ✅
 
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
IPv6 support isn't on by default yet, but it does say 'Support' so I'm inclined to show it (Or just remove this line from the table). Alternatively we could not change anything and wait until we have it on by default. What do you think @mattwoj ?

hyanggi:
As I remembered IPv6 always worked perfectly during the WSL1 days. Are you sure it's not on by default? I can even find discussions online (from that time) talking about how to disable it.

https://askubuntu.com/questions/958876/how-to-disable-ipv6-on-windows-subsystem-for-linux

I do not understand why people these days think IPv6 did not work in WSL1.

hyanggi:
Oh. Do you mean it's not on by default in WSL2? I thought you meant WSL1. Sorry for my misunderstanding.

hyanggi:
IPv6 was always supported and on in WSL1. I can even find online discussions from that time talking about how to disable it.

https://askubuntu.com/questions/958876/how-to-disable-ipv6-on-windows-subsystem-for-linux

It is in WSL2 where IPv6 is not on by default.

Therefore leaving the current table as is (which states WSL1 does not support IPv6) is the worst option.

craigloewen-msft:
merged in! Thank you both for the contribution.

PR #1936, opened 2024-03-19
User/chdhan/wsl troubleshooting doc update 3 18 23

chanpreetdhanjal:
Updated below - 

In [Advanced settings configuration in WSL | Microsoft Learn](https://learn.microsoft.com/en-us/windows/wsl/wsl-config), in the section about "hostAddressLoopback" -> add a note to mention "only IPv4 addresses assigned to the host are supported"

In [Advanced settings configuration in WSL | Microsoft Learn](https://learn.microsoft.com/en-us/windows/wsl/wsl-config), in the section about "localhostForwarding" -> add a note to mention "setting is ignored when networkingMode=mirrored"

In [Accessing network applications with WSL | Microsoft Learn](https://learn.microsoft.com/en-us/windows/wsl/networking#mirrored-mode-networking), update phrase below to also include "IPv6 localhost address ::1 is not supported"
"Connect to Windows servers from within Linux using the localhost address 127.0.0.1"


craigloewen-msft (approved):

keith-horton (approved):

keith-horton (changes requested):
I don't think this statement is correct: "Also only IPv4 addresses assigned to the host are supported." under hostAddressLoopback.

I thought we tested this. I'll need to get back to you.

EDIT: confirmed that the text information provided is correct. Thanks!

keith-horton (approved):
Confirmed: the updated text is 100% accurate.

mattwojo (approved):

PR #1956, opened 2024-05-05
Fix docker run command

FantasqueX:
Current one will get error "/usr/bin/ls: /usr/bin/ls: cannot execute binary file"


mattwojo:
Thanks for the fix @FantasqueX!

PR #1959, opened 2024-05-15
Update wsl-config.md

JoshuaCooper:
Matching the formatting of * near the Key column instead of the value column established lower in the table.


craigloewen-msft:
Great addition! Thank you for changing this.

PR #1966, opened 2024-06-05
Update install.md

hassan20990:


mattwojo:
Thanks for the contribution @hassan20990 , but a bit more detail than what we're looking for in this description.

PR #1972, opened 2024-06-22
Update install.md

squeakyboots:
Additional detail to better ensure the `wsl --set-version` command is successful for users that may have installed WSL prior to WSL2 being available.


craigloewen-msft:
This extra detail is helpful, but can also really bloat our docs as there are lots of areas where we might ask a user to update. Given that `wsl --update` is not *explicitly* needed for this command, I'd rather leave it off of this sentence in the docs. On the WSL product team we're looking at making it easier to update WSL so that users shouldn't have to hit this error in the first place.

Thank you for filing this!

PR #1978, opened 2024-07-08
Extensively clarify "Identify IP address" section.

chjfth:
Previous version text is too simplified. 

Previous version erroneously said `hostname -i`. It should be `hostname -I` actually.


chjfth:

WSL/basic-commands.md
```diff
@@ -168,8 +168,44 @@ To terminate the specified distribution, or stop it from running, replace `<Dist
 
 ## Identify IP address
 
-- `wsl hostname -i` for the IP address of your Linux distribution installed via WSL 2 (the WSL 2 VM address)
-- `cat /etc/resolv.conf` for the IP address of the Windows machine as seen from WSL 2 (the WSL 2 VM)
```
craigloewen-msft, line 172: We should use this command instead:
`ip route show | grep -i default | awk '{ print $3}'`

see here: https://learn.microsoft.com/en-us/windows/wsl/networking#ipv6-access

WSL/basic-commands.md
```diff
@@ -168,8 +168,44 @@ To terminate the specified distribution, or stop it from running, replace `<Dist
 
 ## Identify IP address
 
-- `wsl hostname -i` for the IP address of your Linux distribution installed via WSL 2 (the WSL 2 VM address)
```
craigloewen-msft, line 171: This should be kept simple as just a quick two line explanation, and we should link out deeper info to the networking page.

craigloewen-msft (dismissed):

mattwojo:
Thanks for this contribution @chjfth! Merging!

PR #1996, opened 2024-07-31
make sure the systemd-sysv package is installed

Zhoneym:
> Notice: If your Linux distribution is Debian/Ubuntu/Kali Rolling, you should not only have installed the systemd package, but also make sure the systemd-sysv package is installed.
```bash
sudo apt-get update -y && sudo apt-get install systemd systemd-sysv -y
```

### DIFF
--- a/WSL/systemd.md
+++ b/WSL/systemd.md
@@ -44,7 +44,7 @@ Once your Linux distribution restarts, systemd will be running. You can confirm 
 
 Learn more about [Advanced settings configuration in WSL](wsl-config.md), including the difference between the `wsl.conf` (distribution-specific) and `.wslconfig` (global) config files, how to update automount settings, etc.
 
-> Notice: If your Linux distribution is Debian/Ubuntu/Kali Rolling, you should not only have installed the systemd package, but also make sure the systemd-sysv package is installed.
+If your Linux distribution is Debian/Ubuntu/Kali Rolling, you should not only have installed the systemd package, but also make sure the systemd-sysv package is installed.
 
 ```bash
 sudo apt-get update -y && sudo apt-get install systemd systemd-sysv -y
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
@Zhoneym why is this command needed? Ubuntu and Debian have working systemd installs out of the box, I'm not aware of users needing to run this command to get systemd working.

Zhoneym:
> @Zhoneym why is this command needed? Ubuntu and Debian have working systemd installs out of the box, I'm not aware of users needing to run this command to get systemd working.

@craigloewen-msft  This command is not necessary for WSL2 distributions released on the Microsoft Store because the systemd-sysv package is pre installed in the official distribution. But for custom users, they usually need the deboot tool to generate rootfs (excluding the Linux kernel) and package them into tar files for archiving, and then import them into the wsl2 working environment through the wsl --import command. This makes the systemd-sysv package a mandatory installation, which is not only applicable to Debian or Ubuntu, but also to all various derivative distributions of Debian

WSL/systemd.md
```diff
@@ -44,6 +44,12 @@ Once your Linux distribution restarts, systemd will be running. You can confirm
 
 Learn more about [Advanced settings configuration in WSL](wsl-config.md), including the difference between the `wsl.conf` (distribution-specific) and `.wslconfig` (global) config files, how to update automount settings, etc.
 
+> Notice: If your Linux distribution is Debian/Ubuntu/Kali Rolling, you should not only have installed the systemd package, but also make sure the systemd-sysv package is installed.
```
craigloewen-msft, line 47:
```suggestion
If your Linux distribution is Debian/Ubuntu/Kali Rolling, you should not only have installed the systemd package, but also make sure the systemd-sysv package is installed.
```

craigloewen-msft:
Looks great, thank you for the submission

PR #1999, opened 2024-08-02
Doc updates about resolving .local names

CatalinFetoiu:


keith-horton (approved):

craigloewen-msft (approved):

mattwojo (approved):
Took a crack at adding a bit more context to possibly help folks who may not be as familiar with these concepts.

PR #2000, opened 2024-08-06
Prevent content from being translated in other language documents that should not be translated

zwJimRaynor:
For example in Simplified Chinese documentation the value ”true“ is translated as ”是“ actually it should not be translated https://learn.microsoft.com/zh-cn/windows/wsl/wsl-config#main-wsl-settings


mattwojo:
This is excellent! Thank you for taking the time to contribute this update @zwJimRaynor! Really great fix.

PR #2021, opened 2024-09-15
WSL/wsl-config.md: rewrite lead for clarity

multiplealiases:
This is a rewrite of the lead section, as I feel the previous one was overly long-winded, yet did not meaningfully summarize the most important points of the article in a way that enables readers to quickly answer the questions "what is `wsl.conf`?" and "what is `.wslconfig`?". The information *is* technically present, but it is scattered about in the body of the article when it could be in the lead. 

The main thing this adds is a short comparison table that summarizes the differences between `wsl.conf` and `.wslconfig`, as I feel it will make the summarized information (otherwise totally buried in the sections below) much easier to find for future readers.

I removed a few things I felt could be done away with (in particular the VirtualBox/Hyper-V aside, WSL 2 being a virtual machine, and WSL 1/2 being fundamentally different designs), as well as performing some minor proofreading. It is quite possible I have been too hasty in removing things that were actually necessary.


multiplealiases:

mattwojo:
Thanks for the contribution here and appreciate your attention to detail. We have decided to keep as-is.. part of that decision is that more and more folks are using AI chat to access guidance and tables don't always translate well in that context.

scottdotweb:
That is hands-down the worst response to a documentation patch that I've ever seen.

starchildluke:
More and more folks are using AI chat because it's being forced down their throats without expressing a desire for it. Less accessible documentation is not the way.

dusnm:
Poor decision. Let's keep documentation confusing for humans, because machines are able to parse it better. 

Come on Microsoft, do better.

V0ldek:
This is a late April Fools prank right, there's no way an adult human being wrote that response unironically

S0ulf3re:
> We have decided to keep as-is.. part of that decision is that more and more folks are using AI chat to access guidance and tables don't always translate well in that context.

Please reconsider. It’s called documentation, not AI instructions. I get that you probably have quotas at Microsoft to push AI more into things, but this really isn't the proper way to do that.

woshiwuja:
PR Bad cuz AI cant read :(

oersted:
Even if you agree with the conclusion, is the premise even correct?

LLMs can handle Markdown tables without any issues, for certain LLMs Markdown is almost their "native language" and they are trained extensively to read and write it well, they often output Markdown tables.

Even if this is rendered as HTML, LLMs can handle HTML just as well and there are plenty of tools out there to convert it to Markdown and most training data collection, RAG indexing or retrieval processes processes do that.

In the worst case where it is copy-pasted by a user and mangled quite badly, in my experience LLMs can make sense of that too without issue, particularly such a small table.

richlv:
Could you please reconsider?
This not only makes the documentation much worse, it is also trashing MS image.

User story: As a WSL user, I want helpful documentation, not a word salad to be ingested by LLMs.

MikeTaylor:
After decades of being told, quite rightly "You should write your code first to be comprehensible to humans, and only secondarily for computers", now we have Microsoft telling us the our freaking _documentation_ is supposed to be written for computers instead of humans?

Oh dear or dear.

> Any fool can write code that a computer can understand.  Good programmers write code that humans can understand.
> -- Martin Fowler, _Refactoring_.

andreaslarssen:
Just wow

seangenabe:
> Thanks for the contribution here and appreciate your attention to detail. We have decided to keep as-is.. part of that decision is that more and more folks are using AI chat to access guidance and tables don't always translate well in that context.

Sorry but aren't we supposed to write documentation for humans to understand, not AI?

mattwojo:
Wow, tough crowd. My mistake for adding AI to my reasoning for not accepting this PR. I receive a lot to look through on this content, which is a small part of my job, so when updates seem related to preference for how the info is presented, I tend to spend less time in consideration as it can sometimes seem subjective. Clearly the user community here seems to feel that this is a good update, so I'm happy to take another look. Though I'm also wondering how much is related to my AI comment. 🤷🏽‍♂️☮️ Rest assured humans are my #1 priority for these docs y'all.

timcharper:
It's not a terrible argument, a good share of people _do_ use LLMs to understand docs nowadays. Just... when the LLM gets it wrong, may be nice to have a fallback.

sylveon:
Documentation is written by humans and for humans. If an AI tool can't parse a table, it's a problem with that tool, not the documentation being written.

v-p-b:
> Clearly the user community here seems to feel that this is a good update, so I'm happy to take another look. Though I'm also wondering how much is related to my AI comment.

By reading the comments with my human brain it is obvious to me that the problem raised is your _reasoning_ about why the commit should be rejected (preferring the needs of dubious, sometimes outright harmful technology over actual, intelligent users). I think everyone would just respect your decision if it was underpinned by reasonable arguments.

GossiTheDog:
AI minefield strikes again.

mattwojo:
Going to lock this conversation as this isn't the place to discuss feelings about AI. The best that I can do is reassure you all that I write these docs along with the PM with the intention of being helpful to WSL users. I've learned today that you are sensitive to ensuring human readability. Message received. ❤️ Thanks for your interest in these docs.

mattwojo (approved):

mattwojo:
Appreciate the community feedback on this change. Added in some small edits to make sure that the original information is present in the table and merging the pull request. Thank you for contributing to this repository and making your voice heard.

PyroGenesis:
Thank you for reconsidering your stance. Documentation is for humans, first and foremost, and people do not like playing second fiddle to what AI thinks of it. AI should be molded to be able to understand human-friendly language, not the other way around.

Nerdoholic:
> Thanks for the contribution here and appreciate your attention to detail. We have decided to keep as-is.. part of that decision is that more and more folks are using AI chat to access guidance and tables don't always translate well in that context.

Kindly gfys with your AI-first approach.

I want you scum to come down from your ivory tower and stop fucking over humanity.

LukachoLombardi:
Adapting to arising new technologies is cool and all, but LLM assisted programming is neither advanced nor widespread enough to warrant targeting it first and humans second.

Maybe in 10 years this AI hype will actually have progressed into an actually useful change in common workflow, which would maybe warrant a decision like this. But I guess until then it'll already have learned how to read tables anyway.

zarif98:
Thank god this was approved. Honestly, the dumbest decision I read this year.

PR #2028, opened 2024-09-25
Update WSL shutdown instructions for clarity by specifying PowerShell should be run as administrator before executing the shutdown command.

Civon:
This PR addresses the issue where the first attempt to run wsl --shutdown using an unprivileged PowerShell fails.

## Steps to Reproduce
1. Fill the VHD to its capacity and follow the instructions to fix the read-only fallback error.
1. In the *Read-only fallback error* section, use an unprivileged PowerShell to run `wsl.exe --shutdown`.
1. Observe the error as shown in the image below: 
image

## Solution
Run the command in a privileged PowerShell session.

### DIFF
--- a/WSL/disk-space.md
+++ b/WSL/disk-space.md
@@ -1,7 +1,7 @@
 ---
 title: How to manage WSL disk space
 description: Learn how to check the amount of disk space available, expand the size of the Virtual Hard Disk (VHD), repair a VHD mounting or read-only error, and locate the .vhdx file and disk path for Linux distributions installed with WSL 2.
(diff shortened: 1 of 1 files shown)


Civon:
FYR: I may also need to close Docker Desktop(on Windows) to not getting error _Wsl/Service/CreateInstance/MountVhd/HCS/ERROR_SHARING_VIOLATION_ while running next command `wsl.exe sudo e2fsck -f /dev/sdc`

mattwojo:
Good update - thank you!

PR #2030, opened 2024-09-27
Add note for blocked Microsoft Store

crramirez:


mattwojo (approved):
Thanks for adding this!

PR #2033, opened 2024-09-30
Update networking.md

Horsevad:
Updated that the hostname takes a lowercase -i  as the address flag.  This can be seen by running hostname --help.  

Also made the "under the hood" section more verbose with the full length --ip-addresses command, which might be easier to remember / understand for new users.

### DIFF
--- a/WSL/networking.md
+++ b/WSL/networking.md
@@ -19,12 +19,12 @@ There are two scenarios to consider when identifying the IP address used for a L
 The Windows host can use command:
 
 ```
-wsl -d <DistributionName> hostname -I
+wsl -d <DistributionName> hostname -i
 ```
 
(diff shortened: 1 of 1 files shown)


Horsevad:
Is there any way to get someone assigned to this

mattwojo:
Great update - thank you! (And sorry for the delay -- was busy preparing for Ignite conference)

taoyouh:
This is completely wrong. The lower case "-i" will return a "127.0.1.1" which is a loop back address. The address cannot be accessed from Windows side. However in the document's context it should be the IP address in the NAT network created by WSL (containing the WSL distribution and Windows). The upper-case "-I" will return the outside address of the WSL instance seen by Windows.

You can see a note below in the document:
> Using a lowercase "i" with the hostname command will generate a different result than using an uppercase "I". wsl hostname -i is your local machine (127.0.1.1 is a placeholder diagnostic address), whereas wsl hostname -I will return your local machine's IP address as seen by other machines and should be used to identify the connectaddress of your Linux distribution running via WSL 2.

PR #2034, opened 2024-10-01
Update troubleshooting.md

thiruma0203:
Azure virtual machine does support Nested Virtualization, but cannot use with Trusted Launch. Hence proposing minor correction.


mattwojo:
Thanks for adding this update!

PR #2039, opened 2024-10-12
Add rule to ignore service account's contributions

LiCaoZ:


mattwojo:
Good add! No need to have those service accounts listed as contributors

PR #2040, opened 2024-10-12
Remove the duplicate `feedback_product_url` value

LiCaoZ:


mattwojo:
Why are you trying to remove this feedback link in the settings?

mattwojo:
Oh -- I see it now! Good catch -- no need to repeat that setting. Thanks!!

PR #2043, opened 2024-10-14
Fixing GNOME Text Editor install command

wbsmolen:


crramirez (commented):
Why snap? It forces the prerequisite of having systemd and the snap service installed and running. Basically it is just going to work on Ubuntu out of the box.

PR #2062, opened 2024-11-19
Fix backtick in build-custom-distro.md

CarlosNihelton:


mattwojo:
Thank you @CarlosNihelton for this fix!

PR #2068, opened 2024-11-22
Fix `contributors_to_exclude` did not work

LiCaoZ:


mattwojo:
Awesome - thanks for the fix!!

PR #2083, opened 2024-12-10
Fix incorrect "hostname -i" command in networking.md

taoyouh:
The lower case "-i" will return a "127.0.1.1" which is a loop back address. The address cannot be accessed from Windows side. The upper-case "-I" instead will return the outside address of the WSL instance seen by Windows.

Thus "-I" should be used instead of "-i" to get the host IP seen from WSL. There's already a note saying about this below the changed part.

This commit mainly reverts the change of PR #2033.

### DIFF
--- a/WSL/networking.md
+++ b/WSL/networking.md
@@ -19,12 +19,12 @@ There are two scenarios to consider when identifying the IP address used for a L
 The Windows host can use command:
 
 ```
-wsl -d <DistributionName> hostname -i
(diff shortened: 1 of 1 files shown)


shiinakaze:
@craigloewen-msft May I ask why it has not been merged into the main branch after 3 months ?

craigloewen-msft:
This is a good fix, thank you for the contribution!

PR #2094, opened 2024-12-27
Changed all uppercase word to bold

LucasDondo:


mattwojo:
Thanks for the contribution, but we will handle this sort of subjective formatting decisions.

PR #2098, opened 2025-01-08
Updated docs for server core.

zippy1981:
Based on some experiences I had setting up WSL on server core, both with Azure VMs and Hyper-V on my laptop.

https://serverfault.com/questions/1169719/wsl-distributions-do-not-install-on-windows-server-core-2025/1169879#1169879


zippy1981:
@amitkh-msft Can you take a look at this please?

PR #2102, opened 2025-01-13
Update systemd.md

fcucino:
State that `wsl --version` command could throws an error if outdated and therefore it must be updated to make systemd working on the distribution.


fcucino:

mattwojo:
Great update - thank you!

PR #2131, opened 2025-02-16
Use consistent table formatting in wsl-config

SilentSpud:
Use consistent capitalization and spacing for tables, use `""` as the default for blank strings, use a proper `size` value as a default for `defaultVhdSize`, and add some missing code tags.


SilentSpud:

craigloewen-msft:
Great fix. Thank you for contributing!

PR #2149, opened 2025-03-19
fix: Update `hostname` command with correct flags

AlexDev404:
This updates the `hostname` command's flags in  `networking.md` from the incorrect `--ip-addresses` to the correct `--all-ip-addresses`


AlexDev404:
Fixed by: #2033

PR #2175, opened 2025-05-06
Documented new wsl.conf options

craigloewen-msft:
Added docs for some existing wsl.conf options

### DIFF
--- a/WSL/wsl-config.md
+++ b/WSL/wsl-config.md
@@ -134,6 +134,23 @@ wsl.conf section label: `[boot]`
 | Key | Value | Default | Notes |
 |:----|:----|:----|:----|
 | `command` | string | `""` | A string of the command that you would like to run when the WSL instance starts. This command is run as the root user. e.g: `service docker start`. |
+| `protectBinfmt` | boolean | `true` | Prevents WSL from generating systemd units when systemd is enabled  |
+
+### GPU settings
+
+wsl.conf section label: `[gpu]`
+
+| Key | Value | Default | Notes |
+|:----|:----|:----|:----|
+| `enabled` | boolean | `true` | Allow Linux applications to access the Windows GPU via para-virtualization. |
+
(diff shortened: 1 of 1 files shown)


WSL/wsl-config.md
```diff
@@ -144,6 +144,23 @@ wsl.conf section label: `[boot]`
 | Key | Value | Default | Notes |
 |:----|:----|:----|:----|
 | `command` | string | `""` | A string of the command that you would like to run when the WSL instance starts. This command is run as the root user. e.g: `service docker start`. |
+| `protectBinfmt` | boolean | `true` | Prevents WSL from generating systemd units when systemd is enabled  |
```
crramirez, line 147: does the default value of this is true?

WSL/wsl-config.md
```diff
@@ -144,6 +144,23 @@ wsl.conf section label: `[boot]`
...
+
+### GPU settings
+
+wsl.conf section label: `[gpu]`
+
+| Key | Value | Default | Notes |
+|:----|:----|:----|:----|
+| `enabled` | boolean | `true` | Allow Linux applications to access the Windows GPU via para-virtualization. |
```
crramirez, line 155: does the default value of this is true?

OneBlue: Yes true is the default

OneBlue (approved):

WSL/wsl-config.md
crramirez: so, is this is the current behavior with mesa and d3d12?

WSL/wsl-config.md
crramirez: the comment suggest me that by default it won't generate syste4md units

WSL/wsl-config.md
craigloewen-msft: Yes!

PR #2199, opened 2025-05-22
Fix this saying Linus instead of Linux

benja2998:
How did this typo ship into production?


craigloewen-msft:
Great catch, thank you!

PR #2251, opened 2025-06-25
Fix WSL GitHub repository link text in opensource.md

Mystyle-48:


craigloewen-msft:
Great spot! Thank you for the fix.

PR #2259, opened 2025-07-02
Update install.md - add Windows features as prerequisites

HerrB92:
Hi! I kindly ask to check and then add this information about required Windows features which have to be present/installed, so that the WSL installation does not fail at the "Install Ubuntu" step.

Thank you & Regards!
HerrB92


craigloewen-msft:
When you run `wsl --install` it will install the optional components that you listed, and so they are not required as pre-requisites! Thanks for the feedback.

PR #2260, opened 2025-07-08
Fix variable reference

thefrooh:
Due to a misplaced dollar sign, variable reference was not valid.


craigloewen-msft:
Thank you for the feedback!

PR #2269, opened 2025-07-16
Update wsl-config.md

benhillis:
Remove deprecated pageReporting setting


AndisGrossteins:
> Isn't `pageReporting` here?

Yes but it should be removed from documentation as was done by this PR. The parameter appears to be deprecated and is reported as unknown upon WSL startup in terminal and in Windows Event Log.

PR #2332, opened 2025-09-16
Deprecate the pageReporting setting.

zwJimRaynor:


AndisGrossteins:
There's another instance of the `pageReproting` [in the example code block on line 307](https://github.com/zwJimRaynor/WSL/blob/e4ca31c4c8254820b83264eaaad9388ffc894634/WSL/wsl-config.md?plain=1#L307).

I'm sorry for bothering you instead of making doing my own PR.

PR #2344, opened 2025-09-19
Update case-sensitivity.md for default case option

Red54:
case=off is the new default setting since Build 17763 (1809)
https://docs.microsoft.com/en-us/windows/wsl/release-notes#build-17763-1809


mattwojo:
Thank you for the updates @Red54 !

PR #2405, opened 2025-11-06
Fix typo in networkingMode: net => nat

autozimu:
Looks typo to me. When I tried `net`, WSL complains to me that it's not valid option.


craigloewen-msft (approved):

craigloewen-msft:
It's a typo! Merging, thank you for the feedback :)

Issue #2174, opened 2025-05-03
Fix command for obtaining VM IP address as seen from Windows host


davpaez:
### Type of issue

Code doesn't work

### Feedback

The following article has an error when trying to identify the IP address of the VM as seen from Windows host:
https://learn.microsoft.com/en-us/windows/wsl/networking?source=recommendations#identify-ip-address

The appropriate command to show the VM's IP address as seen by the Windows host should be:
`wsl -d <DistributionName> hostname -I`.

Notice that the correct option is `hostname -I` instead of `hostname -i`

Also, the paragraphs below that command also incorrectly reference the use `hostname -i` and `hostname --ip-addresses` instead of the correct `hostname -I` and equivalent `hostname --all-ip-addresses`

Fortunately, the documentation does NOT repeat the error in the next section called [Accessing a WSL 2 distribution from your local area network (LAN)](https://learn.microsoft.com/en-us/windows/wsl/networking?source=recommendations#accessing-a-wsl-2-distribution-from-your-local-area-network-lan). Instead, there is helpful note clarifying exactly the distinction I'm refering to.

### Page URL

https://learn.microsoft.com/en-us/windows/wsl/networking?source=recommendations#default-networking-mode-nat

### Content source URL

https://github.com/MicrosoftDocs/wsl/blob/main/WSL/networking.md

### Author

@mattwojo

### Document Id

f9e21bfa-6316-7d45-3b34-834a78bcaf0f

### Platform Id

f3658d3a-3068-2c5d-3513-2318700a303d

craigloewen-msft:
Please file this at the github.com/microsoft/wsl repo with logs and we will have the best understanding of what is going on and how we can fix this issue. This repo is just for docs. THank you!

davpaez:
Hello @craigloewen-msft . Thank you for your response.

This is not an issue about a bug in the software. This issue is meant to fix the docs. I'm simply trying to point out the errors that are present in the referenced article.

Please let me know if I should edit the original issue to clarify that. I left it as a "Code doesn't work" type of issue, because that is what the automated system wrote when it created this issue.

davpaez:
@craigloewen-msft I just checked the current version of the article and the error I pointed out has been fixed in most places.

There is still one mention of `hostname -i` in the [IPv6 access](https://learn.microsoft.com/en-us/windows/wsl/networking?source=recommendations#ipv6-access) section that I think should also be corrected to `hostname -I` as well

[...]

PR #2429, opened 2025-12-02
docs: correct hostname flags and command references in networking.md

Shriya-23:
This PR updates the networking documentation to use the correct hostname flags so users receive accurate IP address information in WSL. Closes issue #2174

Changes Made:
Step 1 Replaced the Windows-host command example that used a lowercase -i:
-Old: wsl.exe --distribution <DistroName> hostname -i
-New: wsl.exe --distribution <DistroName> hostname -I

Step 2 Replaced the incorrect long option shown in the under-the-hood explanation:
-Old: hostname --ip-addresses
-New: hostname --all-ip-addresses

Step 3 Updated the IPv6/access section command that referenced the lowercase -i:
-Old: wsl hostname -i
-New: wsl hostname -I

Now it clearly explains that -i shows a placeholder/loopback IP, while -I shows all real IP addresses, so users don’t get confused and accidentally use the wrong command.

Files changed
WSL/networking.md 
Closes Issue
 #2174

### DIFF
--- a/WSL/networking.md
+++ b/WSL/networking.md
@@ -20,12 +20,12 @@ There are two scenarios to consider when identifying the IP address used for a L
 The Windows host can use command:
 
 ```powershell
-wsl.exe --distribution <DistroName> hostname -i
+wsl.exe --distribution <DistroName> hostname -I
 ```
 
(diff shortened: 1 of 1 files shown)


jobayer110 (commented):
Y

PR #2451, opened 2025-12-30
Replace MongoDB references with DocumentDB to prioritize open-source

gahl-levy:
This tutorial lists database options for local development, so recommendations should stick to open-source solutions. MongoDB is source-available under the Server Side Public License (SSPL), which is not an OSI-approved open-source license; see the OSI license list: https://opensource.org/licenses/. This change removes MongoDB as a recommended option and replaces mentions with DocumentDB (with MongoDB compatibility). DocumentDB was developed by Microsoft and is now part of the Linux Foundation, released under the MIT license.  .

### DIFF
--- a/WSL/tutorials/wsl-database.md
+++ b/WSL/tutorials/wsl-database.md
@@ -1,13 +1,13 @@
 ---
 title: Add or connect a database with WSL
-description: Learn how to set up MySQL MongoDB, PostgreSQL, SQLite, Microsoft SQL Server, or Redis on the Windows Subsystem for Linux.
+description: Learn how to set up MySQL DocumentDB, PostgreSQL, SQLite, Microsoft SQL Server, or Redis on the Windows Subsystem for Linux.
 ms.date: 09/18/2023
 ms.topic: get-started
 ---
 
 # Get started with databases on Windows Subsystem for Linux
 
-This step-by-step guide will help you get started connecting your project in WSL to a database. Get started with MySQL, PostgreSQL, MongoDB, Redis, Microsoft SQL Server, or SQLite.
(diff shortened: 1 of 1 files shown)


gahl-levy:
#sign-off

GrantMeStrength:
Thanks for raising this — DocumentDB is a genuinely interesting project and it's good to see it get visibility.

That said, I'd recommend against this change as written, for a few reasons:

**The page is explicitly audience-driven, not license-curated.** The intro cites the Stack Overflow developer survey as the basis for what's listed. MongoDB consistently appears in the top 5 databases on that survey; DocumentDB doesn't appear yet because it's very new. Replacing a tool millions of developers use with one that's largely unknown doesn't serve a developer who arrives here asking 'how do I set up a database in WSL?'

**The SSPL argument would need to apply consistently.** Redis — also listed on this page — switched to SSPL in 2024, which creates the same licensing situation. If SSPL is the disqualifying factor, that's a policy decision that should be applied uniformly across Microsoft docs, not handled file-by-file.

**The replacement removes useful practical guidance.** The current MongoDB section includes a helpful note about systemd support in WSL and the version selector in the MongoDB docs — real gotchas developers hit. The replacement links to a GitHub README. That's a step down in quality for the reader.

**A better approach: add DocumentDB alongside MongoDB.** If the goal is to promote DocumentDB as a modern, MIT-licensed, MongoDB-compatible alternative, the most honest way to do that is to present it as an additional choice — not a silent substitution. Something like:

> *Looking for a fully open-source MongoDB-compatible option? [DocumentDB](https://documentdb.io/) (MIT license, Linux Foundation) offers wire-protocol compatibility and can be used as a drop-in alternative.*

That gives DocumentDB genuine exposure to the right audience (developers already thinking about MongoDB) without misrepresenting what the page covers or removing content that developers depend on.

